PR #36, opened 2011-11-27
Topic/refactor traversal

BiggerNoise:
Incorporates the refactored traversal mechanism.

While I think the separation of concerns in this PR is the right thing for the project, I question if the power that the traversal class brings is really going to be useful enough to offset the additional mental complexity of its functional nature.

I did create a simple interface to traversal that I used for the alter and permissions scripts.  That may strike a better balance between separation of concerns and thought tax.


BiggerNoise:
Nuking. We'll revisit at some point

PR #52, opened 2012-01-27
Configurable Execution Workflow

mpareja:
Our database deployment workflow doesn't quite fit into the default workflow provided by Roundhouse. This pull request adds the ability to customize which folders are executed and when.

A new optional configuration setting (WorkflowConfigFile) allows one to point Roundhouse at a CSV file containing the desired workflow. Here is sample of what the config file would look like for the roundhouse default workflow:

Update, up, onetime
Run First After Update, runFirstAfterUp, onetime
Function, functions, anytime
View, views, anytime
Stored Procedure, sprocs, anytime
Index, indexes, anytime
Run after Other Anytime Scripts, runAfterOtherAnyTimeScripts, anytime

### DIFF
--- a/product/roundhouse.console/Program.cs
+++ b/product/roundhouse.console/Program.cs
@@ -122,7 +122,7 @@
                      option => configuration.DatabaseType = option)
                 // versioning
                 .Add("wfc=|workflowconfig=|workflowconfigfile=",
-                    "WorkflowConfigFile - Custom configuration file listing the folders to execute and the order of execution.",
+                    "WorkflowConfigFile - Custom configuration file listing the folders to execute and the order of execution. Expected format: Friendly Name, Folder Name, Timing. Timing can be 'anytime', 'onetime' or 'everytime'.",
                     option => configuration.WorkflowConfigFile = option)
                 .Add("r=|repo=|repositorypath=",
                      string.Format(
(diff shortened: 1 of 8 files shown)


BiggerNoise:
OK, this P/R has aged nicely, I suppose we can talk about it again. 

I really want to do something like this because I think it will allow us to deprecate and eventually eliminate most of the flags that gunk up the RH CLI. However, I think I'd like to do the file as a JSON file.

If it were an array of objects that had the form:
{"type":"onetime", "directory":"up", "search":"traverse"},
{"type":"onetime", "directory":"runFirstAfterUp", "search":"recurse"},

Search is taking the ideas from https://github.com/chucknorris/roundhouse/pull/80 and pulling them in.

My current plan of attack is:
- 0.8.8: Merge as many P/Rs as possible
- 0.9.0: Modernize Tools & Dependencies
- 0.9.1: Flexible workflow (this P/R)
- 0.9.2: Interactive Mode (https://github.com/chucknorris/roundhouse/issues/270)

Curious if you guys had any thoughts on this.

BiggerNoise:
Closing due to merge conflicts. The idea is sound and I plan to resurrect it as soon as I can clear the rest of the backlog.

PR #64, opened 2012-06-13
Implemented --isuptodate switch

JonAnders:
In our automated deployment scenario for a web application on load balanced web servers, we need to know if there are any changes to the database before running roundhouse. When there are db changes, we need to take take the application offline before updating. When there are only changes to the web app code, we can pull each server out of load balancer, update, and push it back in without any downtime.

To support this scenario, I have implemented the --isuptodate switch. It will simply tell if there are any db changes or not, in a way that is easy to interpret in a script. When combined with the --silent switch, the output is simply "Database is up to date: True" or "Database is up to date: False".

### DIFF
--- a/product/roundhouse.console/Program.cs
+++ b/product/roundhouse.console/Program.cs
@@ -39,6 +39,10 @@
                 {
                     run_diff_utility(set_up_configuration_and_build_the_container(args));
                 }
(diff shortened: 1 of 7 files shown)


BiggerNoise:
Thanks! This will be in the 0.8.8 release.

PR #75, opened 2012-08-29
Dryrun

carllindelof:
Implemented dryrun feature for DefaultDatabaseMigrator


BiggerNoise:
Pulled in #111 instead, but thank you very much.

PR #80, opened 2012-09-07
Refactor of the /searchallinsteadoftraverse option and added ablility to specific an abritrary file ordering.

icedtoast:
Refactor of the /searchallinsteadoftraverse logic into a FileLocator interface.

New /scriptorder option which chooses between traverse, recurse or the specified ordering locators.

Specific Ordering allows you to place a file inside the migration directory which specifies the order of the files containing in that directory and it's subdirectories. This can be used to:
- order scripts in terms of their dependencies
- help handle the merging of database branches.


BiggerNoise:
Again, I thank you for this contribution and I suspect that I'll be incorporating these ideas as time goes on. Unfortunately, because we let this sit so long, the conflicts are going to make it difficult to merge.

That's on us, and I apologize.

PR #82, opened 2012-09-10
Change the StatementSplitter to handle combinations of string literals, comments and batch seperators

icedtoast:
Change RoundHouse to handle mixing of the following for sql server:
- Single line comments `--`
- Multi line comments `/* newline */`
- Nested comments `/* /* nested */ */`
- String literals `'hello there'`
- Batch seperators `GO`

Changes to roundhouse include:
- Added Unit Tests for the regression introduced by [a297dd4](https://github.com/chucknorris/roundhouse/commit/a297dd4) and simliar examples.
- Refactored the Statement Splitter logic into an interface.
- Changed Unit Tests to check on the Split sql, not the replacement of terminators with a "split" string.
- Implemented a tokenizer based splitter for SQL Server.

The SqlServerStatement splitter uses a regular expression to just tokenize the sql. An outer loop then only breaks the sql on "go" statements that are outside of comments or string literals. This has the added advantage of not introducing whitespace into sql batches.

Only SQL Server database targets are affected by these changes

### DIFF
--- a/product/roundhouse.databases.sqlserver/SqlServerDatabase.cs
+++ b/product/roundhouse.databases.sqlserver/SqlServerDatabase.cs
@@ -14,7 +14,15 @@ namespace roundhouse.databases.sqlserver
 
         public override string sql_statement_separator_regex_pattern
         {
-            get { return @"(?<KEEP1>^(?:[\s\t])*(?:-{2}).*$)|(?<KEEP1>/{1}\*{1}[\S\s]*?\*{1}/{1})|(?<KEEP1>'{1}(?:[^']|\n[^'])*?'{1})|(?<KEEP1>\s)(?<BATCHSPLITTER>GO)(?<KEEP2>\s)|(?<KEEP1>\s)(?<BATCHSPLITTER>GO)(?<KEEP2>$)"; }
+            get { throw new NotSupportedException("Use sql_splitter instead"); }
+        }
+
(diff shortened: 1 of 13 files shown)


BiggerNoise:
I know it's been a really long time, but does this P/R need to be merged? Do the changes to the splitter made by @mpareja mean that we can just close this? Or are there other splits covered by this P/R?

icedtoast:
Hi @BiggerNoise ,

This pull-request robustly detects all cases where the sql needs to be split and splits it in a way that doesn't require searching for replacement strings. It handles comment nesting, string literals, basically anything that is valid T-SQL.

I'm not familiar with roundhouse anymore and haven't used it for about 4 years, so I'm uncertain whether you are still using a regex-based splitter. If you are, then yes it would be worthwhile merging this; or at the very least grabbing the extra tests that have been added in this PR.

BiggerNoise:
Thanks for the reply. That's a good idea, I'll start with the tests and see where that takes me.

BiggerNoise:
Thank you for the contribution, but I'm putting this one on the shelf for the time being. I think there's a large number of cases you're covering that are very worthwhile. Unfortunately, with intervening changes, the merge conflicts are are very non-trivial.

When I manage to get things in a bit better shape, I may come back to see if I can extract the core of the ideas. Until then, I am going to close this. Thanks again for the contribution and I apologize for our dropping the ball on merging this.

PR #84, opened 2012-09-10
Store filenames as relative paths

icedtoast:
Update to  https://github.com/chucknorris/roundhouse/pull/78

When used with /storescriptrelative it changes the Migration Runner 
to identify files by their relative path to the migration folder, instead of their filename.
This allows files with the same name to exist in different subdirectories of a migration
folder. Before this change RoundhousE only considered the filename when
checking if it had run the file previously.

e.g. Having a directory layout like this:
up
|- feature1
| |- 001.sql
|
|- feature2
|- 001.sql

Will now run successfully by roundhouse, previously it would identify
'feature2\001.sql' as being a modified copy of 'feature1\001.sql'.

Log messages and the script_name column in the ScriptsRun and
ScriptsRunErrors now store the script_name like 'feature1\001.sql'
instead of '001.sql'.

WARNING:
If you run this on a database already managed by RoundhousE that has
scripts inside subfolders of the migration folders it will re-run those
scripts, because only the filename portion was stored by versions of
RoundHouse before this commit, so only use this option on new database or
when you have moved all of the previously run files out of subfolders.


BiggerNoise:
This is another great idea and I thank you for the contribution. Unfortunately, after merging it in, I am unable to get the feature to work as advertised.

The switch works and the control logic works, but the `file_path.Substring(containingFolder.folder_full_path.Length)` statement in `get_file_name_from()` just returns the file name. 

This is a really great idea, and I look forward to incorporating it when I get into revising the how the workflow works. I am going to close this P/R because I expect that the code will be significantly different when I finally get to working this idea in.

PR #111, opened 2013-04-26
implementing baseline and dryrun if statements

bigminer:
I believe i fixed the line ending

Supersedes https://github.com/chucknorris/roundhouse/pull/97


ferventcoder:
I noticed that dry run has several Pull Requests out on it #75 #111 #124 #189

PR #118, opened 2013-07-18
Upgrade to NHibernate 3.3.3.4000

derjabkin:
- Updated to NHibernate 3.3.3.4000 and log4net 1.2.11
- Removed duplicate prerelease nuget version generation.


BiggerNoise:
Thank you for this P/R. 

Unfortunately, there's quite a bit going here which makes it difficult to tell what change corresponds to what. If you want to resubmit these P/Rs as individual items, I'd be happy to pull them in.

PR #124, opened 2013-09-24
[Enhancement] DryRun and Baseline implementation for RH

kolupaev:
Hi.
I have implemented baseline and dry-run for RH, based on your MockDatabase implementation skeleton.

However, I've re factored it into 3 separate classes:
 * DatabaseDecoratorBase - proxy to proxied database implementation with required methods marked as virtual
 * DatabaseBaselineDecorator - ignores everything, except script insertion
 * DatabaseDryRunDecorator - ignores everything

There are some exceptions happening when --dryrun or --baseline are executed on a non-existing database, however proper implementation of those do require changes in core Database class, which I haven't done yet. 

Also, it seems that integration tests were using incorrect path to SQL scripts folder, I've fixed that as well at https://github.com/kolupaev/roundhouse/commit/b96de52127c8b323fcc4cedc50ecab2954d39af6.

--
Best regards, 
Val


BiggerNoise:
Do we want to go ahead and close this one? The refactoring was nice, but #111 is merged, and implements the baseline and dry run.

ferventcoder:
Yeah, the refactoring was nice, but the one we merged followed the original intent. We may at some point want to go through and refactor things at some point, but I would hope once there is better test coverage.

ferventcoder:
Thanks @kolupaev - we went forward with #111. Sorry this drug out for so long.

PR #128, opened 2013-11-22
Upgrade projects and nugets to dot net 4 where applicable

Mike-Hardy:
In order to incorporate roundhouse into our continuous build and deployment process we need to overcome differing versions of log4net dll. Our product uses the .net 4 version and roundhouse uses .net 2 version


BiggerNoise:
Thanks for this and sorry about ghosting you for so long.

While I am sure that you have long since addressed your issue, here's my plan: I am going to get as many P/Rs merged into the 0.8.x line before I bump .NET versions. That will start the 0.9.x release line.

PR #138, opened 2014-01-31
Pass version via command line argument

### LINKED ISSUE #120: [Enhancement] Pass the version number in the command line
It looks like I have to create some sort of artifact (file) in order to version my db globally. Why cant I just pass this a parameter of the rh.exe ?

alexandermlharris:
*NB: This is the same changelist as Pull Request #135, I've just resubmitted it via it's own feature branch (originally submitted to master - I'm new to git).*

Issue #120 asks for the ability to pass the version via the command line. I had the same requirement (for use in a TeamCity CI environment) to save having to write the managed build number to a file as a separate step.

I've implemented it this PR as follows:

* Add a new Property `Value` to the configuration objects.
* Add new command line argument and description, mapped to this property.
 * `-v|-version`, Optional, but overrides version file switches when set.
* Add a new `CommandLineVersionResolver` which is constructed with this property and checks if it's null / empty.
* Updated the logic in the `VersionResolverBuilder` skip the use of the file based resolvers if the command line resolver is valid.

I've built the project via the build.bat file and tested the produced rh.exe and it works as expected:

* `-v|-version` not **set** - *existing behaviour*
* `-v|-version`  **set** -  *uses version from `-v`*
* `-v|-version` **set**, `-vf|-versionfile` **set** - *uses version from `-v`, doesn't run file checks.*

Please take a look and let me know if you've any questions or comments. I've tried to match the style of the project, but in particular my description may not be to your liking. This is my first time properly running through the github pull process etc. so if I've omitted any steps, or failed to include any required files or changes, my apologies.

Cheers,

Alexander

Closes #120

### DIFF
--- a/product/roundhouse/runners/RoundhouseMigrationRunner.cs
+++ b/product/roundhouse/runners/RoundhouseMigrationRunner.cs
@@ -76,7 +76,7 @@ namespace roundhouse.runners
 
             if (run_in_a_transaction && !database_migrator.database.supports_ddl_transactions)
             {
-                Log.bound_to(this).log_a_warning_event_containing("You asked to run in a transaction, but this dabasetype doesn't support DDL transactions.");
+                Log.bound_to(this).log_a_warning_event_containing("You asked to run in a transaction, but this databasetype doesn't support DDL transactions.");
                 if (!silent)
                 {
                     Log.bound_to(this).log_an_info_event_containing("Please press enter to continue without transaction support...");
--- a/product/roundhouse.console/Program.cs
+++ b/product/roundhouse.console/Program.cs
@@ -127,6 +127,9 @@ namespace roundhouse.console
                      string.Format(
                          "RepositoryPath - The repository. A string that can be anything. Used to track versioning along with the version. Defaults to null."),
                      option => configuration.RepositoryPath = option)
+                .Add("v=|version=",
+                     "Version - Specify the version directly instead of looking in a file. If present, ignores file version options.",
+                     option => configuration.Version = option)
                 .Add("vf=|versionfile=",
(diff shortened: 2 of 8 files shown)


ferventcoder:
@ChrisMissal @BiggerNoise this looks like a great candidate for 0.8.7

ChrisMissal:
Nice addition @alexandermlharris! This looks good to me, any objections to merging this in @ferventcoder?

alexandermlharris:
@ChrisMissal - thanks! Really glad if it helps. :)

product/roundhouse/runners/RoundhouseMigrationRunner.cs
```diff
@@ -76,7 +76,7 @@ public void run()
 
             if (run_in_a_transaction && !database_migrator.database.supports_ddl_transactions)
             {
-                Log.bound_to(this).log_a_warning_event_containing("You asked to run in a transaction, but this dabasetype doesn't support DDL transactions.");
+                Log.bound_to(this).log_a_warning_event_containing("You asked to run in a transaction, but this databasetype doesn't support DDL transactions.");
```
ferventcoder: Nice catch on typo

ferventcoder:
@alexandermlharris would you mind squashing this down to one commit? After that we can move forward with the merge. Thanks!

alexandermlharris:
Sorry, I missed this comment earlier. Do you still need me to do anything?
I've not worked much with git, so any pointers (pun not intended) would be
appreciated! :D

ChrisMissal:
I don't mind taking care of it, unless you wanted to @alexandermlharris?

alexandermlharris:
Sure, go for it. I don't think my git foo is good enough probably! I think
there's a git --rebase in there, or a git --reset or something, but I've
never done it before! :D

alexandermlharris:
Ah, see it's done already. Thanks :)

PR #150, opened 2014-05-27
[Enhancement] Add support for ConfigurationFile parameter

mmooney:
This change introduces a new command line parameter called ConfigurationFile (or configfile or cf), which can point to a simple JSON file.

If the path to a configuration file is provided, it will read properties from the JSON file, and if they match the available configuration values (in ConfigurationPropertyHolder interface), will will copy those values in to the configuration.

If a value does not exist in the configuration file, it will use the default or provided command line value.

If a value exists in the configuration value that does not exist in the ConfigurationPropertyHolder interface, it will be ignored.  

If a value exists but cannot be converted to the correct type (via Convert.ChangeType), it will display an error message.

This JSON works:

``` json
{
    DatabaseName: "TestConfigFile",
    ServerName: "(local)",
    ThisFieldWillBeIgnored: false,
}
```

This JSON will cause a conversion error indicating that "kabook" cannot be converted to a Boolean value for the Restore field:

``` json
{
    DatabaseName: "TestConfigFile",
    ServerName: "(local)",
    ThisFieldWillBeIgnored: false,
    Restore: "kaboom"
}
```

### DIFF
--- a/product/roundhouse.console/Program.cs
+++ b/product/roundhouse.console/Program.cs
@@ -17,6 +17,8 @@ using roundhouse.infrastructure.filesystem;
 using roundhouse.migrators;
 using roundhouse.resolvers;
 using roundhouse.runners;
+using Newtonsoft.Json;
+using System.Collections.Generic;
 
 namespace roundhouse.console
 {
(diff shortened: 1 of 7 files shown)


ChrisMissal:
@mmooney @ferventcoder how are we doing on this one?

PR #154, opened 2014-06-19
Schema bound views support

tsubik:
In our application we have many schema bound views depends on each other. To alter that kind of view there is a need to drop binding first. Views scripts are preprocessed  and if there are dependent schema bound views, binding is dropped and after view is altered, schema binding is restored.

### DIFF
--- a/db/SQLServer/TestRoundhousE/views/vw_SchemaBound.sql
+++ b/db/SQLServer/TestRoundhousE/views/vw_SchemaBound.sql
@@ -1,0 +1,15 @@
+DECLARE @Name VarChar(100)
+DECLARE @Type VarChar(20)
+SET @Name = 'vw_SchemaBound'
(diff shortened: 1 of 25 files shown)


BiggerNoise:
Thank you for the contribution. Given the extremely long time that it's taken to give you a reply, you have no doubt solved this problem.

I have two issues:
1. I think this is getting outside the scope of roundhouse. While, I'm not sure exactly how I would solve this using the tools that RH gives me, I am pretty sure that I wouldn't expect a tool that manages scripts run to solve the problem form.
2. More practically, there's been so many changes to the code base, that the P/R would need to be updated.

I'm going to close this P/R. I am very willing to find out that I am wrong on the scope thing. If you think that this really is something that RH should handle, then I would suggest opening an issue where we can talk about it.

PR #169, opened 2014-12-18
Use a "natural" sort order when enumerating the scripts to execute

iainnicol:
Right now you have to zero-prefix the script names, for example
0001_Foo.sql
0002_Bar.sql
...
0010_Baz.sql

That's slightly ugly.  Worse, the limit might eventually be reached.

Using a natural sort order lets you name the scripts:
1_Foo.sql
2_Bar.sql
10_Baz.sql

### DIFF
--- a/product/roundhouse.tests/infrastructure/NaturalStringComparerSpecs.cs
+++ b/product/roundhouse.tests/infrastructure/NaturalStringComparerSpecs.cs
@@ -1,0 +1,81 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.IO;
+
+namespace roundhouse.tests.infrastructure
+{
+    using roundhouse.infrastructure;
+    using bdddoc.core;
+    using developwithpassion.bdd.contexts;
+    using developwithpassion.bdd.mbunit;
+    using developwithpassion.bdd.mbunit.standard;
+    using developwithpassion.bdd.mbunit.standard.observations;
+
+    public class NaturalStringComparerSpecs
+    {
+        public abstract class concern_for_comparer : observations_for_a_sut_with_a_contract<IComparer<string>, NaturalStringComparer>
+        {
+            protected static object result;
+
+            context c = () => { };
+        }
+
+        [Concern(typeof(NaturalStringComparer))]
+        public class when_sorting : concern_for_comparer
+        {
+            private const string zeroOne = "001_Foo";
+            private const string zeroTwo = "002_Foo";
+            private const string zeroThree = "003_Foo";
+            private const string zeroFour = "004_Foo";
+            private const string zeroTen = "010_Foo";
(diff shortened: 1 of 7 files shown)


iainnicol:
I think this fell off everybody's radar (mine included) over the holidays.

Shall we merge this then?  Or if not, let's just close it.  Regards.

product/roundhouse/infrastructure/NaturalStringComparer.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// &quot;1_Foo&quot; &lt; &quot;2_Foo&quot; &lt; &quot;10_Foo&quot;.
+    /// </summary>
+    public sealed class NaturalStringComparer : IComparer<string>
+    {
+        [DllImport("shlwapi", CharSet=CharSet.Unicode)]
+        private static extern int StrCmpLogicalW(string psz1, string psz2);
+
+        int IComparer<string>.Compare(string x, string y)
+        {
+            return StrCmpLogicalW(x, y);
```
ferventcoder: The only downside to using this is that it won't work in *nix. Is there a way to provide a pure .NET solution to this problem?

product/roundhouse.tests/infrastructure/NaturalStringComparerSpecs.cs
```diff
@@ -0,0 +1,66 @@
...
+            [Observation]
+            public void numbers_should_sort_correctly()
+            {
+                string[] unsorted = new[] { one, ten, eleven, two, three };
+                string[] sortedExpected = new[] { one, two, three, ten, eleven };
+                string[] sortedActual = unsorted.OrderBy(x => x, new NaturalStringComparer()).ToArray();
+                sortedActual.should_be_equal_to(sortedExpected);
+            }
+        }
+    }
```
ferventcoder: @iainnicol still waiting on a spec for `zeroOne` compared with `one` to know which one comes first.

ferventcoder: I see you've added `two`, which is good, but not quite my intent.

ferventcoder:
@iainnicol I believe you are correct. I think it fell off of everyone's radar. I commented on a couple of things that I believe are still holding this up. We have folks right now that are trying to get RH working with Mono, so this natural sort order at the very least would need a pure .NET fallback if not on Windows platform.

iainnicol:
Thanks for the updated review comments.

I'd agree portability is important, and obviously more important than this.  I'll close this (as a WONTFIX).  If I find the time to make this portable I'll reopen it. But I don't want to leave this pull request open because in honesty I might not get round to it.

ferventcoder:
Fair enough. Someone may end up picking this up and making the adjustments.

PR #179, opened 2015-03-01
Fix for bug #176: The creation of database in SQL 2000 does not work.

### LINKED ISSUE #176: [BUG] The creation of database in SQL 2000 does not work
In the methode 

```
roundhouse.databases.sqlserver2000.SqlServerDatabase.create_database_script
```

these code doesn't work in SQL 2000.

```
public override string create_database_script()
{
	return string.Format(
		@"
			DECLARE @Created bit
			SET @Created = 0
			IF NOT EXISTS(SELECT * FROM sys.databases WHERE [name] = '{0}')
			BEGIN
				CREATE DATABASE [{0}]
				SET @Created = 1
			END
			SELECT @Created
			",
		database_name);
}
```

We should use **sysdatabase** without the dot.

sicil1ano:
This is a PR for the bug #176 .The fix uses the function @@VERSION of SQL Server to get the information about the current MS SQL Server instance (as explained also here: http://support2.microsoft.com/default.aspx?scid=kb;en-us;321185), and the function PATINDEX (https://technet.microsoft.com/en-us/library/aa276881%28v=sql.80%29.aspx) to check the specific pattern ('Microsoft SQL Server 2000'). In case the pattern is found (then PATINDEX returns an int > 0) and the database doesn't exist, the script creates the requested database. Let me know if you have other suggestions and/or edits.

This closes #176.

### DIFF
--- a/product/roundhouse.databases.sqlserver2000/SqlServerDatabase.cs
+++ b/product/roundhouse.databases.sqlserver2000/SqlServerDatabase.cs
@@ -93,7 +93,6 @@ namespace roundhouse.databases.sqlserver2000
 
         public override string create_database_script()
         {
-            //Using global variable @@VERSION to get the information about current MS SQL Server instance.
             return string.Format(
                 @"
                         DECLARE @Created bit
--- a/product/roundhouse.databases.sqlserver2000/SqlServerDatabase.cs
+++ b/product/roundhouse.databases.sqlserver2000/SqlServerDatabase.cs
@@ -93,15 +93,19 @@ namespace roundhouse.databases.sqlserver2000
 
         public override string create_database_script()
         {
+            //Using global variable @@VERSION to get the information about current MS SQL Server instance.
             return string.Format(
                 @"
                         DECLARE @Created bit
                         SET @Created = 0
-                        IF NOT EXISTS(SELECT * FROM sys.databases WHERE [name] = '{0}') 
-                         BEGIN 
-                            CREATE DATABASE [{0}] 
-                            SET @Created = 1
-                         END
+                        IF(PATINDEX('%Microsoft SQL Server 2000%',@@VERSION) > 0)
+                        BEGIN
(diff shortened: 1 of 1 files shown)


product/roundhouse.databases.sqlserver2000/SqlServerDatabase.cs
```diff
@@ -93,15 +93,19 @@ public override void run_database_specific_tasks()
...
             return string.Format(
                 @"
                         DECLARE @Created bit
                         SET @Created = 0
-                        IF NOT EXISTS(SELECT * FROM sys.databases WHERE [name] = '{0}') 
-                         BEGIN 
-                            CREATE DATABASE [{0}] 
-                            SET @Created = 1
-                         END
+                        IF(PATINDEX('%Microsoft SQL Server 2000%',@@VERSION) > 0)
```
ferventcoder: What is this call? I see the note above but I still don't know what you are doing here. And what happens when the number is 0? I don't see a call to error?

pfillion: In my view, no need to check the version of SQL Server. Anyway, this code is in the specialized class SQL2000 and it will never call for other server types.

ferventcoder: That's kind of what I was thinking here.

sicil1ano: Hello guys! Sorry for my late answer, but I had some health problems. Yes, I didn't realize immediately that the code was specialized for SQL Server 2000 (I believe I had a slow moment..), and I'm sorry for that. :( I removed the useless parts in the code (comment and version check).

ferventcoder:
I missed that you updated this. Any chance you can squash down those commits to just one? 

On amending your commit, would you mind using the summary `(GH-176) Fix Sql 2000 database creation`

ferventcoder:
Superseded by #196

ferventcoder:
Thanks for the contribution! This should be out in 0.8.7. :)

PR #180, opened 2015-03-04
fixes Searchallinsteadoftraverse bug

### LINKED ISSUE #172: Searchallinsteadoftraverse with trailing '='-sign
In the console application the code that adds the option searchallinsteadoftraverse is defined with a trailing '='-sign. This can cause the next argument to not be parsed successfully. I'm not sure why this works sometimes, but I had the problem when passing the the argument with an argumentlist trough powershell.

ChrisMissal:
This resolves #172


ferventcoder:
:+1:

PR #181, opened 2015-03-04
only retry on sql connection errors

### LINKED ISSUE #173: Wrong error message for script error when running in transaction
When RoundhousE is running in transaction mode and a script fails the real error message is lost and a strange error i s showed instead. For instance:
```
CREATE TABLE #TEST (TestKey int NOT NULL)
GO
INSERT #Test(TestKey) VALUES ('123A')
```
Would get an error saying that the table `#Test` does not exist, but the real error is that `'123A'` cannot be converted to `int`.

This is probably due to the "really naive retry logic" in AdoNetDatabase.cs (currently :132). Probably the transaction is rolled back before the retry. Maybe it could be considered to change the code to throw the original exception in the case where both `run_command_with(...)` fail. Like:

```
//really naive retry logic. Consider Lokad retry policy 
//this is due to sql server holding onto a connection http://social.msdn.microsoft.com/Forums/en-US/adodotnetdataproviders/thread/99963999-a59b-4614-a1b9-869c6dff921e 
try 
{ 
    run_command_with(sql_to_run, connection_type, parameters); 
} 
catch (Exception ex) 
{ 
    Log.bound_to(this).log_a_debug_event_containing("Failure executing command, trying again. {0}{1}", Environment.NewLine, ex.ToString()); 
    try
    {
        run_command_with(sql_to_run, connection_type, parameters);
    }
    catch (Exception new_ex)
    {
        //If failing re-throw the original exception.
        throw ex;
    }
} 
```

An other solution would be to look at the SqlException's error number to only handle the main issue that the retry solves like:

```
//really naive retry logic. Consider Lokad retry policy 
//this is due to sql server holding onto a connection http://social.msdn.microsoft.com/Forums/en-US/adodotnetdataproviders/thread/99963999-a59b-4614-a1b9-869c6dff921e 
try 
{ 
    run_command_with(sql_to_run, connection_type, parameters); 
} 
catch (SqlException ex) 
{ 
    Log.bound_to(this).log_a_debug_event_containing("Failure executing command, trying again. {0}{1}", Environment.NewLine, ex.ToString()); 
    if (ex.Number == 233)
    {
        run_command_with(sql_to_run, connection_type, parameters);
    }
    else
    {
        //Re-throw the original exception.
        throw;
    }
} 
```

ChrisMissal:
this will prevent any other errors from being hidden. fixes #173.

### DIFF
--- a/product/roundhouse/databases/AdoNetDatabase.cs
+++ b/product/roundhouse/databases/AdoNetDatabase.cs
@@ -7,11 +7,13 @@ namespace roundhouse.databases
     using System.Collections.Generic;
     using System.Data;
     using System.Data.Common;
+    using System.Data.SqlClient;
     using connections;
     using parameters;
 
     public abstract class AdoNetDatabase : DefaultDatabase<IDbConnection>
     {
+        private const int sql_connection_exception_number = 233;
         private bool split_batches_in_ado = true;
 
         public override bool split_batch_statements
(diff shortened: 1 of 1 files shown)


ferventcoder:
+1

PR #189, opened 2015-05-12
Don't run SQL if /dryrun flag set

jordanwallwork:


ferventcoder:
I thought maybe there was another PR already out for this?

ferventcoder:
There are actually about 4

jordanwallwork:
Have any been pulled then? It's a nuisance having to make this change whenever I update

BiggerNoise:
OK. This is silly, let's pick one and merge it. #189 has the virtue of simplicity, just fixing the bug,

#111 and #124 also implement baseline. Do we want that as well? #124 is a more extensive refactoring, but it looks like it starts to move the logic in a better direction. It's starting to get tangled in there.

@ferventcoder - I'm dusting off my C# chops (hide the women and children). I don't have super strong feelings about any of these. If you or @ChrisMissal want to weigh in, just assign one of these to me and I will verify it and merge it for the weekend.

BiggerNoise:
Pulled in #111  instead, but thank you very much.

jordanwallwork:
Great news, thanks

PR #191, opened 2015-06-09
Migrated roundhouse.databases.oracle to managed Oracle ODP.NET driver

Buthrakaur:
There's no need to install Oracle Client on the machine nor to switch between 32/64bit versions of the driver with managed Oracle provider.


ferventcoder:
So is the requirement to move to .net 4.0 to take advantage of odp.net? If so, please split this up to multiple branches / pull requests and state that one depends on the other. Then we can evaluate the upgrade to .net 4.0 first separately. And then move to this. 

Also, don't use master branch for PR's. You get weird things like your last commit that is there based on pulling instead of git fetch and rebase. 

Hopefully this is helpful to get you started on where you need to go to move this forward.

Buthrakaur:
The main concern is using managed Oracle client, which doesn't require Oracle Client software installation. I'll clean up the PR a try it again - thanks for guidance.

I have a problem with the buildscript - ILMerge part in particular. I added Oracle.ManagedDataAccess.dll, but it doesn't get into the ILMerged folder so the merge process fails. Do you have any hint what should I modify in the buildscript to make it work?

ferventcoder:
Yes, in the build.custom folder look for ilmerge.pre.build.

ferventcoder:
Actually that is an incorrect statement. Just make sure that the file falls into the correct folder. It should do this as long as copy assembly is true.

PR #194, opened 2015-06-24
Add drop database support for Azure Database

vansha:
Rh fails to remove database, when running against existing azure database using `-drop` switch.

1. `msdb.dbo.sp_delete_database_backuphistory` is not supported in Azure Database
   The error message is following:
```
EXEC msdb.dbo.sp_delete_database_backuphistory @database_name = 'Roundhouse'
Msg 40515, Level 15, State 1, Line 25
Reference to database and/or server name in 'msdb.dbo.sp_delete_database_backuphistory' is not supported in this version of SQL Server.
```

2. `ALTER DATABASE DbName SET SINGLE_USER WITH ROLLBACK IMMEDIATE` is not
   needed in Azure Database, because `DROP DATABASE DbName` works well
   without it regardless of users being connected. While this statement
   when executed on Azure Database results with following error:
```
Msg 42008, Level 16, State 5, Line 1
ODBC error: State: 42000: Error: 1468 Message:'[Microsoft][ODBC Driver 11 for SQL Server][SQL Server]The operation cannot be performed on database "Roundhouse" because it is involved in a database mirroring session or an availability group. Some operations are not allowed on a database that is participating in a database mirroring session or in an availability group.'.
Msg 5069, Level 16, State 3, Line 1
ALTER DATABASE statement failed.
```

### DIFF
--- a/product/roundhouse.databases.sqlserver/SqlServerDatabase.cs
+++ b/product/roundhouse.databases.sqlserver/SqlServerDatabase.cs
@@ -205,15 +205,20 @@ namespace roundhouse.databases.sqlserver
         public override string delete_database_script()
         {
             return string.Format(
-                @"USE master 
-                        IF EXISTS(SELECT * FROM sys.databases WHERE [name] = '{0}' AND source_database_id is NULL) 
-                        BEGIN 
+                @"USE master
+                        DECLARE @azure_engine INT = 5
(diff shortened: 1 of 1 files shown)


vansha:
Hi guys! Any chances you could look at this pull request?

We're currently using Roundhouse in production to manage SQL Azure database migrations.  Everything works smoothly except the `/drop` database feature (we're recreating databases on dev environments in the cloud). While it's not a critical feature, since database could be removed using Azure API, it's handy to leave that work for the RoundhousE.

ferventcoder:
This looks good to me, do you want to take a look @ChrisMissal ? If you are good with it, feel free to merge.

Issue #178, opened 2015-02-18
Research Framework Upgrade


ChrisMissal:
See what the upgrade path looks like and create any issues necessary for getting to 4.0 or 4.5. As well as Visual Studio solution version.

BiggerNoise:
Would it make sense to look at upgrading the Visual Studio version at the same time?  

Work doesn't give me Visual Studio any more (I work in *nix land these days), so it would be nice to move this to the 2013 Community Edition.

ChrisMissal:
Yes indeed :+1:

ChrisMissal:
I'll get it running on 2013 Community and get those changes in a PR

ChrisMissal:
Looks like NAnt isn't too happy with the framework changes. Is this also the right time to replace it? @ferventcoder @BiggerNoise

ferventcoder:
Might just need fixes? I may have seen this mentioned in one other place. Any replacement needs to be xplat and provide the same functionality.

ChrisMissal:
Yeah, I'm digging in. I really don't want to have to replace it if I can help it. A small part of me just doesn't want to have to trudge through this much xml. :wink:

ChrisMissal:
Sorry for jumping the gun @ferventcoder @BiggerNoise, ignore my NAnt comments. I was developing in new VM that appears to have some framework inconsistencies.

PR #195, opened 2015-07-01
Upgrade from VS2010 .NET 3.5 to VS2013 .NET 4.5

ChrisMissal:
The only thing that wasn't super straightforward was the need to upgrade ILMerge to address a bug in v4.0+.


ChrisMissal:
Closes #178 

Can I get a quick look @ferventcoder or @BiggerNoise?

ferventcoder:
LGTM

ILMerge has a key you can pass to it to ensure .NET 4 platform:

https://github.com/chocolatey/choco/blob/master/.build.custom/ilmerge.replace.build#L28

`/targetplatform:v4`

ferventcoder:
It looks like you may need to address that as well in the following places:
* https://github.com/chucknorris/roundhouse/blob/master/build.custom/ilmerge.replace.build#L30-L31
* https://github.com/chucknorris/roundhouse/blob/master/build.custom/ilmergeDLL.build#L34-L35

ferventcoder:
branches in the upstream.... :/

PR #196, opened 2015-07-02
(GH-176) Fix Sql 2000 database creation

### LINKED ISSUE #176: [BUG] The creation of database in SQL 2000 does not work

ChrisMissal:
Squashed commits from @sicil1ano's master branch per @ferventcoder's request.

Fixes #176, closes #179.

### DIFF
--- a/product/roundhouse.databases.sqlserver2000/SqlServerDatabase.cs
+++ b/product/roundhouse.databases.sqlserver2000/SqlServerDatabase.cs
@@ -97,11 +97,11 @@ namespace roundhouse.databases.sqlserver2000
                 @"
                         DECLARE @Created bit
                         SET @Created = 0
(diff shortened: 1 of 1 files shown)


ferventcoder:
Sweet!

PR #200, opened 2015-07-08
Improve MSBuild error handling for 'RecoveryMode'

StevenLiekens:
This PR adds a reference to Microsoft.Build.Utilities.v4.0 for its `TaskLoggingHelper` class.

Code that previously threw `ArgumentOutOfRangeException` now instead log an MSBuild error.

MSBuild errors that were logged using the `TaskLoggingHelper` now cause the Roundhouse task to exit early and crash the build script (unless `ContinueOnError=true`).

### DIFF
--- a/product/roundhouse.tasks/Roundhouse.cs
+++ b/product/roundhouse.tasks/Roundhouse.cs
@@ -7,16 +7,27 @@
     using infrastructure.app.logging;
     using infrastructure.containers;
     using infrastructure.filesystem;
-    using infrastructure.logging;
+
     using Microsoft.Build.Framework;
     using migrators;
+
+    using Microsoft.Build.Utilities;
+
     using resolvers;
     using runners;
     using Environment = environments.Environment;
+    using Logger = roundhouse.infrastructure.logging.Logger;
 
     public sealed class Roundhouse : ITask, ConfigurationPropertyHolder
     {
         private RecoveryMode recoveryMode;
+
+        private readonly TaskLoggingHelper loggingHelper;
+
+        public Roundhouse()
+        {
+            this.loggingHelper = new TaskLoggingHelper(this);
+        }
 
         #region MSBuild
 
@@ -30,7 +41,7 @@
         bool ITask.Execute()
         {
             run_the_task();
-            return true;
+            return !this.loggingHelper.HasLoggedErrors;
         }
 
         #endregion
@@ -148,7 +159,7 @@
                 }
                 else
                 {
-                    this.Logger.log_an_error_event_containing("The value of 'RecoveryMode' must be one of these values: 'NoChange', 'Simple' or 'Full'. Actual value was '{0}'.", value);
+                    this.loggingHelper.LogError("The value of 'RecoveryMode' must be one of these values: 'NoChange', 'Simple' or 'Full'. Actual value was '{0}'.", value);
                 }
             }
         }
@@ -176,6 +187,11 @@
 
         public void run_the_task()
         {
+            if (this.loggingHelper.HasLoggedErrors)
+            {
+                return;
+            }
+
             Log4NetAppender.configure_without_console();
             ApplicationConfiguraton.set_defaults_if_properties_are_not_set(this);
 
--- a/product/roundhouse.tasks/Roundhouse.cs
+++ b/product/roundhouse.tasks/Roundhouse.cs
@@ -63,7 +63,7 @@
 
         public string RunAfterCreateDatabaseFolderName { get; set; }
 
-		public string RunBeforeUpFolderName { get; set; }
+        public string RunBeforeUpFolderName { get; set; }
 
         public string UpFolderName { get; set; }
 
@@ -142,12 +142,14 @@
             set
             {
                 RecoveryMode result;
-                if (!Enum.TryParse(value, true, out result))
+                if (Enum.TryParse(value, true, out result))
                 {
-                    throw new ArgumentOutOfRangeException("value", value, "The value of 'RecoveryMode' must be one of these values: 'NoChange', 'Simple' or 'Full'.");
+                    this.recoveryMode = result;
                 }
-
-                this.recoveryMode = result;
+                else
+                {
+                    this.Logger.log_an_error_event_containing("The value of 'RecoveryMode' must be one of these values: 'NoChange', 'Simple' or 'Full'. Actual value was '{0}'.", value);
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


product/roundhouse.tasks/Roundhouse.cs
```diff
@@ -142,12 +153,14 @@ public string RecoveryMode
...
                 RecoveryMode result;
-                if (!Enum.TryParse(value, true, out result))
+                if (Enum.TryParse(value, true, out result))
                 {
-                    throw new ArgumentOutOfRangeException("value", value, "The value of 'RecoveryMode' must be one of these values: 'NoChange', 'Simple' or 'Full'.");
+                    this.recoveryMode = result;
+                }
+                else
+                {
+                    this.loggingHelper.LogError("The value of 'RecoveryMode' must be one of these values: 'NoChange', 'Simple' or 'Full'. Actual value was '{0}'.", value);
```
ferventcoder: We should already be using an MSBuild Logger - so just logging the error should work - https://github.com/chucknorris/roundhouse/blob/master/product/roundhouse/infrastructure/logging/custom/MSBuildLogger.cs

StevenLiekens: How do I get a reference to the logger? I looked at the IoC container stuff that is already built in, but couldn't make much sense of it.

ferventcoder: Plus this may not be only called with MSBuild, so you can't put MSBuild specifics here.

product/roundhouse.tasks/Roundhouse.cs
```diff
@@ -30,7 +41,7 @@ public sealed class Roundhouse : ITask, ConfigurationPropertyHolder
         bool ITask.Execute()
         {
             run_the_task();
-            return true;
+            return !this.loggingHelper.HasLoggedErrors;
```
ferventcoder: Getting to this may be harder without direct access to the logger. It should be possible, just not quite as straightforward.

product/roundhouse.tasks/Roundhouse.cs
```diff
@@ -174,6 +187,11 @@ public string RecoveryMode
 
         public void run_the_task()
         {
+            if (this.loggingHelper.HasLoggedErrors)
+            {
+                return;
+            }
```
ferventcoder: Also not always MSBuild coming through here, so probably better not to explicitly use this call here.

product/roundhouse.tasks/Roundhouse.cs
ferventcoder: Actually there may be a simpler way. Just have a local boolean called property_sets_errored and set it to true if any properties errored. You may also want to hold a logging message until the logger is configured.

ferventcoder: Look at https://github.com/chucknorris/roundhouse/pull/200/files#diff-0a45691fae4571c0a1ed7b44894a05c2R202 for a convention to setting the error appropriately

ferventcoder: More in that it is looking to see whether some values were set.

StevenLiekens: I'm starting to get confused. Is this class used by anything besides unit tests and MSBuild itself?

ferventcoder: Yes, this is both MSBuild and NAnt. 

The simpler way is to set a string to an error if you are unable to set and then write the error out. Boom, no need for adding msbuild tasks and taking over the loggers.

ferventcoder: The simpler way is to set a string to an error if you are unable to set and then write the error out in the execute task. Boom, no need for adding msbuild tasks and taking over the loggers.

StevenLiekens: But then you don't get cool things like line/column numbers of the error in your build script, which is what the `TaskLoggingHelper` does.

ferventcoder: Would you want it to continue on error? I wanted simple recovery, but whoops I did something wrong. Now what state is the db in?

product/roundhouse.tasks/Roundhouse.cs
StevenLiekens: That's not what ContinueOnError does. It only controls whether failed tasks cause the build script to crash. I don't know for sure about this, but I think that uncaught exceptions cause the script to crash regardless of the ContinueOnError setting.

product/roundhouse.tasks/Roundhouse.cs
StevenLiekens: That's not what ContinueOnError does. It only controls whether failed tasks cause the build script to crash. I don't know for sure about this, but I think that uncaught exceptions cause the script to crash regardless of the ContinueOnError setting.

ferventcoder: Well in the case of this code updates, if ContinueOnError is set, it will be set to `NoChange` (the default) - https://github.com/chucknorris/roundhouse/blob/master/product/roundhouse/databases/RecoveryMode.cs#L5.

ferventcoder: FWIW, to move forward, at this time I don't want the logginghelper or msbuildtasks referenced. We can take a look at what is set during execute (after everything is loaded) and throw an error at that time like we already have a convention for.

StevenLiekens: Besides breaking NAnt compatibility, why would you not want to use MSBuild specific classes?

ferventcoder: Wow, forgot how long RH has been around. I think my biggest issue is more the loggingHelper.

ferventcoder:
Overall, I think I am probably fine with these changes considering being reminded this is just MSBuild and no longer NAnt as well.

ferventcoder:
The one thing that is still out is whether the ILMerge works without issue since we've added a new reference.

StevenLiekens:
If NAnt is no longer a concern, then ideally I would want to change this class to extend `Microsoft.Build.Utilities.Task` instead of implementing `ITask` and wiring everything up manually.

ferventcoder:
Baby steps. Let's get this in and then move to the bigger refactoring

ferventcoder:
Last review calls for @drusellers @ChrisMissal @BiggerNoise

ChrisMissal:
:+1:

ferventcoder:
@StevenLiekens thanks for your contribution! If you want to follow up on a longer ticket for the changes you mentioned, let's do that. Create an issue so you can reference that issue in your commits. :)

PR #203, opened 2015-07-21
Handle files with leading batch splitter

### LINKED ISSUE #202: "Could not find stored procedure 'GO'." error for script starting with GO
While converting from Tarantino I got stuck for a bit with this error because a script started with `GO`. I'm also able to reproduce with a script that contains only `GO`.

dahlbyk:
I believe this fixes #202.

I also applied an equivalent change to handle leading splitters in the other databases, but not sure if that's appropriate?

/cc @ChrisMissal

### DIFF
--- a/product/roundhouse.databases.oracle/OracleDatabase.cs
+++ b/product/roundhouse.databases.oracle/OracleDatabase.cs
@@ -17,7 +17,7 @@ namespace roundhouse.databases.oracle
 
         public override string sql_statement_separator_regex_pattern
         {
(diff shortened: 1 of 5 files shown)


ferventcoder:
Any chance you can prefix those commits with `(GH-202)`?

ferventcoder:
Also perhaps a bit of a commit body explaining the change.  

I imagine you are going to end up squashing these two down to one commit anyway. :)

dahlbyk:
More betterer?

ferventcoder:
Better - curious if `^` is start of line in multi-line regex? I think it likely is if you've confirmed this works appropriately. :)

ChrisMissal:
Wonderful, thanks @dahlbyk! I'm good to merge unless you have anything else @ferventcoder?

ferventcoder:
Nada, I'm :+1:

PR #134, opened 2014-01-27
Patch 1

alexandermlharris:
Fixed a small typo on some of the warning text.

PR #204, opened 2015-07-22
(GH-120) Pass version number in the command line

ChrisMissal:
- Add CommandLineVersionResolver.cs which extracts version directly from a
    string.
  - Update VersionResolverBuilder.cs to include the new resolver. Logic
    now groups the file based resolvers together and skips them if the
    command line resolver is valid.
  - Added Version property to the configuration objects.
  - Fix small typo / spelling error (already submitted in Pull Request #134

### DIFF
--- a/product/roundhouse.console/Program.cs
+++ b/product/roundhouse.console/Program.cs
@@ -138,6 +138,9 @@ namespace roundhouse.console
                      string.Format(
                          "RepositoryPath - The repository. A string that can be anything. Used to track versioning along with the version. Defaults to null."),
                      option => configuration.RepositoryPath = option)
+                .Add("v=|version=",
+                     "Version - Specify the version directly instead of looking in a file. If present, ignores file version options.",
+                     option => configuration.Version = option)
                 .Add("vf=|versionfile=",
                      string.Format("VersionFile - Either a .XML file, a .DLL or a .TXT file that a version can be resolved from. Defaults to \"{0}\".",
                                    ApplicationParameters.default_version_file),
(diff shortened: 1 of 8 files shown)


ChrisMissal:
This squashes all the commits with a the original issue in the commit message.

Closes #120, closes #135, closes #138, and closes #204.

vansha:
Just wanted to say thank you @alexandermlharris @ChrisMissal @ferventcoder for this very handy feature!

alexandermlharris:
@vansha Happy to contribute! We haven't actually got around to using this in anger believe it or not, though we may be using it again soon. Still, very happy if this helps others!

ferventcoder:
@ehexter would be happy to know this went in as well.

ferventcoder:
@erichexter ^^

erichexter:
lol, we actually just retired our use of roundhouse. :) it did the job for what we needed, but we have worked ourselves out of the frameworks that cannot manage their own migrations.

ferventcoder:
> frameworks that cannot manage their own migrations

@erichexter that's interesting - last I knew RH was able to handle migrating parts of itself on upgrade. Unless you mean as a migration framework, of which RH is definitely not. Just an engine to run sql files (or sql migrations). It works with things like NHibernate/Entity Framework to convert migrations to sql files. I've also never gotten back to it, but a Fluent Migrations integration bit is also very doable.

erichexter:
@ferventcoder what I mean by this specifically, was this. We were using roundhouse to manage a bunch of migrations for some Entity Framework Model first stuff that I inherited. We added RH to manage the migrations, then we added code first to implement new models. At that point we ran both RH and EF migrations together, Once we converted all the Model First entities to Code First Entities, we were able to just us the EF Migration tool. This process took about a year, because we were not in a hurry to convert because we had RH to take the migration pain away from the model first stuff.

ferventcoder:
@erichexter oh right on. No worries. :) I still use it even with Code First

PR #208, opened 2015-08-04
postgres dispose issue workaround

worldspawn:


worldspawn:
Creates a postgre (isolated) only solve for the dispose problem. I got a meeting now havent had a chance to test it

ChrisMissal:
I like that this is isolated to just Postgres. I don't like that it swallows an exception that I feel should be handled by the base class. I also don't like that I don't know more about Postgres to offer a better suggestion.

Thanks for this @worldspawn. Can I get your thoughts please @ferventcoder?

ferventcoder:
:+1: Merge away. :)

PR #217, opened 2015-09-15
The MD5 hashing algorithm has been replaced with SHA1.

mroutt:
The MD5 hashing algorithm is not FIPS compliant. I have updated the project to use the SHA1 hashing algorithm so that it can be used where Windows has been configured to not allow MD5.

### DIFF
--- a/product/roundhouse.tests/cryptography/CryptographicServiceSpecs.cs
+++ b/product/roundhouse.tests/cryptography/CryptographicServiceSpecs.cs
@@ -15,35 +15,35 @@ namespace roundhouse.tests.cryptography
             [Test]
             public void this_is_me_learning_hashing()
             {
-                MD5 dude = MD5.Create();
+                SHA1 dude = SHA1.Create();
                 string text_to_hash = "I want to see what the freak is going on here";
                 byte[] clear_text_bytes = Encoding.UTF8.GetBytes(text_to_hash);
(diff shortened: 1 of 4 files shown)


ferventcoder:
Thanks. The point of MD5 hashing is for checksumming, not for security. We should definitely look into allowing different algorithms, they may just take up more space and be slower. By default though, MD5 should still be the algorithm.

mroutt:
I completely understand that the MD5 algorithm is not used for encryption and really shouldn't be a security issue. It was simply much easier for me to switch to SHA1 than write up an explanation for the auditors. Your point about MD5 taking up less space is a good one. For 99% of users, MD5 is probably the way to go. Perhaps my next submission will simply be a flag that defaults to MD5.

ferventcoder:
@mroutt I was hoping you would add to this. :)

As it stands, it's a good request, but it's not complete.

BiggerNoise:
Thanks for this contribution. I have recently merged in another P/R that provides an implementation of MD5 in FIPS compliant environments. I don't know if that will allow you to avoid your auditor's concerns (since the machine wont need an MD5 implementation).

For my part, since roundhouse already boasts a staggering number of flags, I would need to understand why our implementation of MD5 does not fill the requirement. If it doesn't, I would really like to understand why.

For now, I am closing this. If it turns out that there really is a need for this, then we'll figure out a plan of action.

PR #186, opened 2015-04-09
Mono compatibility

s-tarasov:

PR #224, opened 2015-10-26
Pg enhancements

### LINKED ISSUE #223: Can find the requested .Net Framework Data Provider
```
roundhouse.databases.postgresql.PostgreSQLDatabase with provider Npgsql does not provide a facility for creating a database at this time.
Unable to find the requested .Net Framework Data Provider.  It may not be installed.
RoundhousE encountered an error.
System.ArgumentException: Unable to find the requested .Net Framework Data Provider.  It may not be installed.
   at System.Data.Common.DbProviderFactories.GetFactory (System.String providerInvariantName) in <filename unknown>:line 0
   at roundhouse.databases.AdoNetDatabase.GetAdoNetConnection (System.String conn_string) in <filename unknown>:line 0
   at roundhouse.databases.AdoNetDatabase.open_connection (Boolean with_transaction) in <filename unknown>:line 0
   at roundhouse.migrators.DefaultDatabaseMigrator.open_connection (Boolean with_transaction) in <filename unknown>:line 0
   at roundhouse.runners.RoundhouseMigrationRunner.run () in <filename unknown>:line 0
Unable to find the requested .Net Framework Data Provider.  It may not be installed.
System.ArgumentException: Unable to find the requested .Net Framework Data Provider.  It may not be installed.
   at System.Data.Common.DbProviderFactories.GetFactory (System.String providerInvariantName) in <filename unknown>:line 0
   at roundhouse.databases.AdoNetDatabase.GetAdoNetConnection (System.String conn_string) in <filename unknown>:line 0
   at roundhouse.databases.AdoNetDatabase.open_connection (Boolean with_transaction) in <filename unknown>:line 0
   at roundhouse.migrators.DefaultDatabaseMigrator.open_connection (Boolean with_transaction) in <filename unknown>:line 0
   at roundhouse.runners.RoundhouseMigrationRunner.run () in <filename unknown>:line 0
```

drusellers:
This is a series of commits that i needed to make postgres work on the Mac OS using Mono.

Closes #223


product/roundhouse.console/app.config
```diff
@@ -1,5 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
+  <system.data>
+    <DbProviderFactories>
+      <add name="Npgsql" invariant="Npgsql"
+      description=".Net Framework Data Provider for Postgres"
+      type="Npgsql.NpgsqlFactory, Npgsql, Version=3.0.3.0, Culture=neutral, PublicKeyToken=5d8b90d52f46fda7"/>
+    </DbProviderFactories>
+  </system.data>
```
ferventcoder: What is this? How does it affect other databases?

product/roundhouse.databases.postgresql/PostgreSQLDatabase.cs
```diff
@@ -88,7 +88,15 @@ public override void set_provider()
...
-            return string.Format("Server={0};Database={1};Port=5432;UserId={2};Password={3};", server_name, database_name, user_name, password);
+            var csb = new NpgsqlConnectionStringBuilder();
+            
+            csb.Host = server_name;
+            csb.Database = database_name;
+            csb.UserName = user_name;
+            csb.Password = password;
+            csb.Port = 5432;
+
+            return csb.ToString();
```
ferventcoder: :+1:

product/roundhouse.console/app.config
drusellers: This is what is required to use https://github.com/chucknorris/roundhouse/blob/master/product/roundhouse/databases/AdoNetDatabase.cs#L31. We are telling `System.Data.DbProviderFactories` that we have a factory called `Npgsql` (which comes from https://github.com/chucknorris/roundhouse/blob/master/product/roundhouse.databases.postgresql/PostgreSQLDatabase.cs#L86). Npgsql implements the required interface.

To my knowledge, this would need to be done for all other database types. You can see something similar for MySql here - https://github.com/chucknorris/roundhouse/blob/master/product/roundhouse.databases.mysql/MySqlDatabase.cs#L81

Oracle: https://github.com/chucknorris/roundhouse/blob/master/product/roundhouse.databases.oracle/OracleDatabase.cs#L87

Some can be registered in the machine config by installers, in this case I'm just adding it for just this application.

product/roundhouse.databases.postgresql/PostgreSQLDatabase.cs
drusellers: At some point we may want to add the port number to the list of configuration elements.

ferventcoder: Yeah, definitely.

ferventcoder: https://github.com/chucknorris/roundhouse/issues/225

product/roundhouse.console/app.config
ferventcoder: Would it be possible to put it in the class itself? If not totally understand. I think there was a reason I left it out of the config itself. I think the purpose was to try to be as agnostic as possible in the console/msbuild/lib.

ferventcoder: To my knowledge, if this value is here, it won't work with roundhouse.lib, which will only have roundhouse.dll.

drusellers: `DbProviderFactories` only supports get methods. `GetFactory` and `GetFactoryClasses`. It looks like you would have to get down and dirty to add programmatically http://stackoverflow.com/questions/1117683/add-a-dbproviderfactory-without-an-app-config

drusellers: Seeing the same commit over in #186 for MySQL - we will probably want to build something like is in the above link via a separate pull request and then we can DRY all of that up.

drusellers:
@ChrisMissal @ferventcoder what's left that is blocking this from coming in. Looking to get PGSQL + Mono rocking.

ferventcoder:
This could be good to go. If you are comfortable with it, merge it and do some real testing with it. The worst that could come of it is that we have to back out the changes.

PR #232, opened 2015-11-16
{WIP}(GH-235) Add support for multiple environments

### LINKED ISSUE #235: Support Multiple Environments
RoundhousE should allow you to set multiple environments
(e.g. --environments=NEWZEALAND,CUSTOMER).

As an example, we want to support country specific features and also customer specific features. This allows us to use the same set of RoundhousE files, then set the environments depending on where we are deploying to and which customers we are supporting.

If we have the following files:
0000 - main.sql
0001 - new zealand feature.NEWZEALAND.env.sql
0002 - australia feature.AUSTRALIA.env.sql
0003 - customer feature for any country.CUSTOMER.env.sql

Then script 0000, 0001 and 0003 will run, but not the Australia file.

cjzness:
Allows you to set multiple environments (e.g. _--environments=NEWZEALAND,CUSTOMER_).

If we have the following files:
0000 - main.sql
0001 - new zealand feature.NEWZEALAND.env.sql
0002 - australia feature.AUSTRALIA.env.sql
0003 - customer feature for any country.CUSTOMER.env.sql

Then script 0000, 0001 and 0003 will run, but not the Australia file.

Scripts are checked to see if they contain any environment from the environment set given (matching the current behaviour), remaining compatible with current use cases (i.e. https://groups.google.com/d/topic/chucknorrisframework/x2CAAWLgJ9I/discussion) and nothing needs to change for people who don't want to use multiple environments.

Three environment flags have been added to match the current 3:
Current: env, environment, environmentname
Added: envs, environments, environmentnames

Closes #235

### DIFF
--- a/product/roundhouse.tests/migrators/DefaultDatabaseMigratorSpecs.cs
+++ b/product/roundhouse.tests/migrators/DefaultDatabaseMigratorSpecs.cs
@@ -22,10 +22,10 @@ namespace roundhouse.tests.infrastructure.containers
         public abstract class concern_for_database_migrator : observations_for_a_sut_without_a_contract<DefaultDatabaseMigrator>
         {
             protected static object result;
-            protected static DefaultEnvironment environment;
+            protected static DefaultEnvironmentSet environment_set;
 
             context c = () => {
-                            environment = new DefaultEnvironment(new DefaultConfiguration {EnvironmentName = "TEST"});
+                            environment_set = new DefaultEnvironmentSet(new DefaultConfiguration {EnvironmentNames = "TEST"});
                         };
         }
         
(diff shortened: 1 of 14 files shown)


product/roundhouse.tests/migrators/DefaultDatabaseMigratorSpecs.cs
```diff
@@ -37,31 +37,31 @@ public class when_determining_if_we_are_in_the_right_environment : concern_for_d
...
+                sut.this_is_an_environment_file_and_its_in_the_right_environment("BOBTEST.something.ENV.sql", environment_set).should_be_false();
             }  
             
             [Observation]
             public void if_given_BOBTEST_in_the_middle_and_in_TEST_should_return_false()
             {
-                sut.this_is_an_environment_file_and_its_in_the_right_environment("something.BOBTEST.ENV.sql", environment).should_be_false();
+                sut.this_is_an_environment_file_and_its_in_the_right_environment("something.BOBTEST.ENV.sql", environment_set).should_be_false();
             }
         }
```
ferventcoder: Leave the original tests and add new ones.

product/roundhouse/consoles/DefaultConfiguration.cs
```diff
@@ -37,7 +37,7 @@ public sealed class DefaultConfiguration : ConfigurationPropertyHolder
         public string VersionTableName { get; set; }
         public string ScriptsRunTableName { get; set; }
         public string ScriptsRunErrorsTableName { get; set; }
-        public string EnvironmentName { get; set; }
+        public string EnvironmentNames { get; set; }
```
ferventcoder: This looks like it could potentially break existing usages of Migrate.

ferventcoder: Like DropkicK and other tools that use the API.  You need to provide a backwards compatible EnvironmentName as well.

product/roundhouse/infrastructure.app/ConfigurationPropertyHolder.cs
```diff
@@ -37,7 +37,7 @@ public interface ConfigurationPropertyHolder
         string VersionTableName { get; set; }
         string ScriptsRunTableName { get; set; }
         string ScriptsRunErrorsTableName { get; set; }
-        string EnvironmentName { get; set; }
+        string EnvironmentNames { get; set; }
```
ferventcoder: Same goes here. This will break existing API users. You must continue to provide EnvironmentName (and possibly give a deprecation warning.

ferventcoder:
This is a good start. Please file an issue with this problem statement, then update your commit to point to that issue. Something like `(GH-###) Support Multiple Environments` and then a body that explains the problem set and what it solves. Your commit currently contains almost no information about this change and why it is necessary. 

Thanks in advance! Let us know if you need any help or pointers on how to do the above.

cjzness:
Thanks for the feedback. I have created an issue and referenced it from the commit.

To remain compatible with EnvironmentName, should I be creating a separate run_sql method in DatabaseMigrator/DefaultDatabaseMigrator which takes an EnvironmentSet while keeping the existing run_sql method signature (using Environment)?

ferventcoder:
> To remain compatible with EnvironmentName, should I be creating a separate run_sql method in DatabaseMigrator/DefaultDatabaseMigrator which takes an EnvironmentSet while keeping the existing run_sql method signature (using Environment)?

I'm not sure that is necessary. Just convert the single one to an environment set and pass it through.

cjzness:
Sorry for the delay.

I have updated my commit. It now uses the single environment for the original environment tests and I have added more tests to verify the checks work correctly when running with multiple environments. I have also added back EnvironmentName for API compatibility (with a deprecation warning).

Please let me know if I have misunderstood/missed something.

Issue #237, opened 2015-11-20
Occasional "Timeout expired" SqlException on retrieving list of executed scripts when migrating Azure Sql Database


vansha:
We experience following exception occasionally when migrate database on Azure Sql Database:
```
2015-11-20 16:00:19,541 [INFO ] - --------------------------------------------------
2015-11-20 16:00:19,542 [INFO ] - Looking for Update scripts in "C:\project\DatabaseScripts\scripts\20.up". These should be one time only scripts.
2015-11-20 16:00:19,542 [INFO ] - --------------------------------------------------
2015-11-20 16:00:22,513 [WARN ] - 0001.InitialSchema.sql - I didn't find this script executed before.
NHibernate.Exceptions.GenericADOException: could not execute query
[ SELECT this_.id as id1_0_, this_.version_id as version2_1_0_, this_.script_name as script3_1_0_, this_.text_of_script as text4_1_0_, this_.text_hash as text5_1_0_, this_.one_time_script as one6_1_0_, this_.entry_date as entry7_1_0_, this_.modified_date as modified8_1_0_, this_.entered_by as entered9_1_0_ FROM RoundhousE.ScriptsRun this_ ]
[SQL: SELECT this_.id as id1_0_, this_.version_id as version2_1_0_, this_.script_name as script3_1_0_, this_.text_of_script as text4_1_0_, this_.text_hash as text5_1_0_, this_.one_time_script as one6_1_0_, this_.entry_date as entry7_1_0_, this_.modified_date as modified8_1_0_, this_.entered_by as entered9_1_0_ FROM RoundhousE.ScriptsRun this_] ---> System.Data.SqlClient.SqlException: Timeout expired.  The timeout period elapsed prior to completion of the operation or the server is not responding. ---> System.ComponentModel.Win32Exception: The wait operation timed out
   --- End of inner exception stack trace ---
   at System.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParserStateObject.ReadSniError(TdsParserStateObject stateObj, UInt32 error)
   at System.Data.SqlClient.TdsParserStateObject.ReadSniSyncOverAsync()
[...]

PR #238, opened 2015-11-20
(GH-237) Increase timeout for NHibernate commands

vansha:
This is to prevent Timeout Expired exceptions when migration is running
on a slow/high latency server (the case for Azure Sql Server) or a
database with a lot of update scripts recorded in a `ScriptsRun` table.

The timeout of 5 minutes is currently hard-coded. Probably the better
way would be to add an additional configuration option. On the other
hand I'm not sure user should care about that additional config option.

Relates to #237


ferventcoder:
What's the default timeout?

ferventcoder:
I'm pretty much :+1: on this as is. @ChrisMissal @drusellers thoughts?

vansha:
The default command timeout is 30 seconds.

ferventcoder:
Thanks!

PR #242, opened 2016-01-27
Updated MySqlDatabaseConnection string to allow User Variables so SPR…

Draftkings:
Issues Fixed: 
1) In order to allow SPROC files with variables to work, updated initialize_connections and build_connection_string to set 'Allow User Variables' = True
2) Fixed the CommandTimeout commandline option not being respected by MySqlDatabase. Fix involved updating 'run_sql' to use the method 'setup_database_command' instead of doing 
`var connection = connection_type == ConnectionType.Admin
                ? admin_connection.underlying_type().downcast_to<MySqlConnection>()
                : server_connection.underlying_type().downcast_to<MySqlConnection>();`

In order to use 'setup_datatbase_command' I needed to not use the MySqlScript class which only accepts a MySqlConnection object instead of an an IDbCommand (which is where the CommandTimeout is set)

### DIFF
--- a/product/roundhouse.databases.mysql/MySqlDatabase.cs
+++ b/product/roundhouse.databases.mysql/MySqlDatabase.cs
@@ -6,6 +6,9 @@ namespace roundhouse.databases.mysql
     using infrastructure.extensions;
     using infrastructure.logging;
     using MySql.Data.MySqlClient;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Globalization;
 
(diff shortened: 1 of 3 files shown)


ferventcoder:
Is this adding tokens for mysql?

Draftkings:
It allows Tokens for MySQL without using the `MySqlScript` class which then allows the RoundhousE command timeout command line option to work with `MySQL`

Issue #152, opened 2014-06-09
System.IO.FileLoadException with log4net, Version=1.2.11.0


khaledm:
Hi, I'm getting this "Could not load file or assembly 'log4net, Version=1.2.11.0, Culture=neutral, PublicKeyToken=669e0ddf0bb1aa2a' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)" exception. I understand that roundhouse.lib is using  log4net v1.2.10.0. Not sure why it is seeking to load v1.2.11.0. I've received the same exception with an existing application as well as a throw-away console app. Any suggestions?

StevenLiekens:
I just ran into this. Is there anything I can do to fix it for just my project? I'm trying to use the MSBuild `<Roundhouse>` task in an `AfterBuild` target. The task crashes with a `FileNotFoundException`, no matter what version of log4net I'm using.

ChrisMissal:
Can this be solved with a binding redirect?

ferventcoder:
Unfortunately no, because the key the assembly was signed with changed.

ChrisMissal:
Would this be fixed if we upgrade log4net as part of the 0.8.7 NuGet release?

ferventcoder:
As long as it is fixed in the nuspecs, yes.

StevenLiekens:
Resolved in #245

PR #245, opened 2016-04-01
Fix log4net dependency

StevenLiekens:
The NuGet packages depend on an older version of log4net than the version that is actually used throughout the code. This causes installation problems in most NuGet clients.

The lastest version is 1.2.15, roundhouse depends on 1.2.11, but the NuGet packages are configured to require exactly 1.2.10 instead of 1.2.11 or newer.

This PR fixes the version restriction, and makes it less strict because it causes problems in projects that use a newer version of log4net.

### DIFF
--- a/nuget/roundhouse.lib/roundhouse.lib.nuspec
+++ b/nuget/roundhouse.lib/roundhouse.lib.nuspec
@@ -14,7 +14,7 @@
     <tags>roundhouse db migration database migrator chucknorris</tags>
     <iconUrl>https://raw.github.com/chucknorris/roundhouse/master/nuget/RoundhousE_Logo.NuGet.jpg</iconUrl>
     <dependencies>
-      <dependency id="log4net" version="[1.2.10]" />
+      <dependency id="log4net" version="2.0.0" />
       <dependency id="FluentNHibernate" version="1.3.0.717" />
     </dependencies>
   </metadata>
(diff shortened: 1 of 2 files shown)


ferventcoder:
This is a breaking change for clients that are still using the older log4net.  This is only seen when using the library correct?

StevenLiekens:
I don't think roundhouse in its current state can be used with log4net 1.2.10. The code depends on log4net 1.2.11, and I don't think .NET will automatically bind to that version (correct me if I'm wrong).

(please note that the nuget package is version 2.0.0 but the dll version is 1.2.11: https://www.nuget.org/packages/log4net/2.0.0)

StevenLiekens:
An alternate solution is downgrading all the code to log4net 1.2.10 and keeping the dependency on 1.2.10.

In that case, do remove the version restriction. Requiring _exactly_ 1.2.10 instead of 1.2.10 _or newer_ causes more problems than it solves.

ferventcoder:
The problem is that it's either 1.2.10 OR 1.2.11+. That's due to using different keys to sign log4net.

ferventcoder:
> The code depends on log4net 1.2.11

I'm not sure what you mean by this. All the code released this is not true. Was something introduced recently that changed that?

StevenLiekens:
> Was something introduced recently that changed that?

all `packages.config` files in the master branch reference log4net package 2.0.0, which is log4net.dll version 1.2.11

ferventcoder:
Looks like I need to do a blame.

StevenLiekens:
> The problem is that it's either 1.2.10 OR 1.2.11+. That's due to using different keys to sign log4net.

They do still publish log4net with the old key on their website, but only builds with the new public key are published on nuget.org.

I opened the latest released `rh.exe` (0.8.6) in dotPeek and saw that it uses log4net 1.2.11, but it's impossible to tell which public key was used because that information was lost in the ILMerge.

ferventcoder:
Why don't we do this? Make it all uniform for now. Then go back later and figure out why the bump occurred.

StevenLiekens:
FYI log4net was bumped to version 1.2.11 in this commit d21978fa8e3a2569442d56a991e2385e600df0f3

ferventcoder:
Fixes https://github.com/chucknorris/roundhouse/issues/152

Issue #247, opened 2016-04-19
Mixed line endings not ignored when comparing hashes


rfbain:
This is a follow up to issue: Ignore EOL changes when checking whether script has changed.  There is still a problem when the original text has a mix of windows and unix line endings.  An attempt is made to convert the new text to all windows line endings and all unix line endings, but neither one matches the original with mixed line endings.  I'd like to see a command line switch that lets the user choose to set line endings to all windows or to all unix prior to computing the hash which gets saved to table ScriptsRun.  Note, I'm only asking for the hash to be changed to enable future checks to work correctly.  I would like the original version of the text to be saved in column text_of_script.  If others would like the text to be updated as well, perhaps that could use different values for the switch.

ferventcoder:
So I'm not fully following. If we are ignoring the `\r\n` \n` line endings for comparison, what does it matter what it is stored as?

ferventcoder:
As in, I'm pretty sure once the ignore is in, line endings would no longer be a factor.

rfbain:
Perhaps an example will help.  Think of the following being included in a sql script.

-- this line was copied from a unix source.\n
-- this line was copied from a windows source.\r\n

Then someone looks at the file in a windows editor and saves it.  Now it becomes.

-- this line was copied from a unix source.\r\n
-- this line was copied from a windows source.\r\n

Function have_same_hash_ignoring_platform in class DefaultDatabaseMigrator will show the file as having been changed and the script will either be run again, or trigger an error or warning when the script should be skipped.

ferventcoder:
Again, I'm not sure it would be a factor, that's what ignoring the line endings is about. It doesn't matter if the line ending is windows or unix format, or even older Mac format with `\r`, the line endings are completely ignored with the ignore line endings code. If they are not, it's a bug in the code itself.

rfbain:
The enclosed zip file has 4 batch files, and the logs produced by executing those files.  Some sql scripts have lines that consistently end in CrLf.  Some have lines that consistently end in Lf.  And some have lines with a mix of line endings.  You should be able to tell which is which from the filenames.  Batch files CrLf_Lf and Lf_CrLf run as expected.  This shows line endings are ignored when the line endings in individual files are consistent, even if they differ from one file to another.  Batch files CrLf_Mixed and Mixed_Crlf show the bug when line endings differ in individual files.
[rh_issue.zip](https://github.com/chucknorris/roundhouse/files/228457/rh_issue.zip)

[...]

PR #248, opened 2016-04-22
Normalizing line endings prior to computing hash.

rfbain:
Fix to issue #247

### DIFF
--- a/product/roundhouse/cryptography/MD5CryptographicService.cs
+++ b/product/roundhouse/cryptography/MD5CryptographicService.cs
@@ -15,7 +15,14 @@ namespace roundhouse.cryptography
 
         public string hash(string clear_text_of_what_to_hash)
         {
-            byte[] clear_text_bytes = Encoding.UTF8.GetBytes(clear_text_of_what_to_hash);
+            const string line_ending_windows = "\r\n";
+            const string line_ending_unix = "\n";
+            const string line_ending_mac = "\r";
+            string normalized_text_of_what_to_hash = clear_text_of_what_to_hash
+                .Replace(line_ending_windows, line_ending_unix).Replace(line_ending_mac, line_ending_unix);
+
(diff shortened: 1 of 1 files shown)


drusellers:
@ChrisMissal @ferventcoder is this good to come in?

ferventcoder:
It is, but the commit needs a body that really explains what this fixes and why.

ferventcoder:
@rfbain apologies for not taking a look sooner, any chance you can address my comment about having a commit body to go along with that summary? 

For details on how to format that, this is a really good explanation and has links to reasoning behind it - https://github.com/chocolatey/choco/blob/master/CONTRIBUTING.md#prepare-commits

Alternatively @drusellers @ChrisMissal if you want to rebase the commit you can pull it in with the good message.

rfbain:
(GH-247) Normalize Line Endings Prior to Calculating and Saving the Hash

Previously, the hash was calculated based on the line endings in the source file.  The hash was then saved to the database without benefit of line ending normalization.  Code run at a later time to compare the hash of the updated file checked if normalized line endings in the updated file matched the original file.

Without the change, there is a problem when the original file has a mix of windows and unix line endings.  Its hash will not match the hash of the updated file.  The script with changes limited to line ending changes will be run when it should be skipped.

ferventcoder:
@rfbain and this will work with scripts already calculated as well, if I understand correctly.

rfbain:
No.  For scripts that have a mix of line endings and have hash values that were calculated prior to this change, my fix will not solve the problem by itself.  Nor will it make the problem worse than it was before.  I thought about adding code to read the original source code from the database and recalculating the hash, but I did not want to introduce that significant a performance hit to the process.  I think the best option to fix old hash values would be to run roundhouse with the baseline option.

BiggerNoise:
Thank you for the contribution. I have made changes in the database migrator that supersede this P/R. Putting the code there allows checking a number of cases and allows compatibility with previously calculated scripts.

Issue #240, opened 2015-12-14
When reading the Version file the content is not trimmed.


drusellers:
We are writing our version.txt file and it was written with a trailing new line. When RH reads it it takes everything and makes for odd input data.

Please trim the content.

Either here: https://github.com/chucknorris/roundhouse/blob/master/product/roundhouse/resolvers/TextVersionResolver.cs#L34 (My vote if for here)

or here

https://github.com/chucknorris/roundhouse/blob/master/product/roundhouse/infrastructure/filesystem/WindowsFileSystemAccess.cs#L57

ferventcoder:
Definitely the first location. It should be very specific to version

drusellers:
My thoughts as well. :)

macfarmw:
I've added the trim to fix up version when it's read from the file. No problem there. I am stuck on the test. Can anyone can offer a clue on what I have done wrong? I am not very familiar with the bdd framework or Rhino so I may need to do more homework before trying this. 

I've setup the mock FileSystemAccess and I'm attempting to supply the FileSystemAccess and version file path here:
[TestVersionResolverSpecs.cs](https://github.com/macfarmw/roundhouse/blob/Issue240-TrimVersionFileContents/product/roundhouse.tests/resolvers/TextVersionResolverSpecs.cs#L57)

When resolve_version is invoked these mock values are blank.

ferventcoder:
@macfarmw mock means it doesn't actually access a real file - you must supply what the return would have been.

ferventcoder:
I looked at the spec and it looks okay. Are you able to run with debugging?

macfarmw:
No. I just installed TestDriven.Net but can't step through code yet. Thanks for checking. I'll fiddle with this a little longer till I find the mistake.

ferventcoder:
TD.net works well - or at least it used to.

macfarmw:
OK. Got it now. I just forgot to supply a return value for the call to get_full_path in the constructor.

`the_filesystem.Stub(x => x.get_full_path(the_versionfile)).Return(the_versionfile); `

It's working fine now. I'll get this tidied up and ready to pull tomorrow.

PR #252, opened 2016-05-25
(GH-240) Trim version file contents

macfarmw:
Added Trim to TextVerisonResolver class where it reads the version to remove any leading or trailing whitespace that may been added to the file.


ferventcoder:
Closes #240

PR #253, opened 2016-05-27
Add init for tls 1.x support

kgav:
TLS 1.2 connections are failing when a connection is initialised to a database sever running TLS 1.2 only - this init method adds support for these connections.

### DIFF
--- a/product/roundhouse.console/Program.cs
+++ b/product/roundhouse.console/Program.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Linq;
+using System.Net;
 using log4net;
 using roundhouse.consoles;
 using roundhouse.databases;
@@ -26,6 +27,8 @@ namespace roundhouse.console
         private static void Main(string[] args)
         {
             Log4NetAppender.configure();
+
+            init_security_protocol();
 
             int exit_code = 0;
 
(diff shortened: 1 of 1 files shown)


product/roundhouse.console/Program.cs
```diff
@@ -429,6 +432,12 @@ private static void run_diff_utility(ConfigurationPropertyHolder configuration)
             }
         }
 
+        public static void init_security_protocol()
+        {
+            // allow tls
+            ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls | SecurityProtocolType.Tls11 | SecurityProtocolType.Tls12 | SecurityProtocolType.Ssl3;
```
ferventcoder: Does this work in .NET 3.5? .NET 4.0? I thought this was only available once you moved up to .NET 4.5+?

ferventcoder: Just confirmed it's definitely not there in .NET 4.0.

`SecurityProtocolType.Tls11 | SecurityProtocolType.Tls12`

Here is 3.5 - https://msdn.microsoft.com/en-us/library/system.net.securityprotocoltype(v=vs.90).aspx
Here is 4.0 - https://msdn.microsoft.com/en-us/library/system.net.securityprotocoltype(v=vs.100).aspx

ferventcoder: Nevemind - looks like @ChrisMissal already upgraded to .NET 4.5 awhile back - https://github.com/chucknorris/roundhouse/commit/b0990af067edd589b0eee4fd1432ad9016768f7a

Issue #190, opened 2015-06-03
Doc - Coding standards


ferventcoder:
Need to document our coding standards and implement code helpers such as resharper team settings.

ferventcoder:
It's basically everything is lowercase except for classnames and
interfaces. When variables, properties, parameters, methods, etc have
multiple words, you separate with an underscore. The sacred cow "I"
prepended to interfaces is not observed here.

macfarmw:
I have a pretty good start on a ReSharper version 9 DotSettings file. I can finish that and submit it. For controlling indentation spaces vs tabs an editorconfig and Rebracer file for those with either of these tools  installed would work nicely. If agreed I'll get started on this. I don't know of any others. I saw a Visual Studio forum discussion about possible built in support for the .editorconfig format.

ferventcoder:
Thanks @macfarmw

macfarmw:
Pull request #254

PR #254, opened 2016-05-29
(GH-190) Coding standards 

macfarmw:
Add contributing guide that outlines the naming rules and config files for ReSharper and EditorConfig support.


ferventcoder:
This looks good. Can you clean up the commits so there is not a merge commit so we can merge? Thanks!

macfarmw:
I'm happy to do that but I'm not sure I understand what you are asking for.

macfarmw:
I thought I had it but I think I just ended up with the same thing again. I used rebase and squashed the last commits together but then I can't push that back to github unless I pull again first which adds back that merge commit.

macfarmw:
OK. I think I have it. Is this right? I used git rebase -i to squash the recent commits and then git push --force to overwrite the remote repository.

roundhouse
```diff
@@ -0,0 +1 @@
+Subproject commit 72c618d9f19ae7fe79d22f7a8fb37c32a2fc4077
```
ferventcoder: What is this? :D

ferventcoder:
It all looks good, I just have a question about the roundhouse file that looks like a submodule.

macfarmw:
Yikes! Here the heck did that come from? I'll have to fix that and squash again. It's OK I clearly need some practice on that anyway.

PR #261, opened 2016-08-02
TableCreationScripts for PostgreSQL do not handle schema name properly.

tsaluszewski:
Updated postgreSQL related scripts to handle schema correctly. Tested in my postgres environment and worked as a charm! :)


hmvs:
Would be nice if it will be merged.

Issue #243, opened 2016-02-16
Unit and integration tests - are they used?


erikbra:
I have a WIP issue #187, but I would love to be able to run some tests before trying to submit a pull request again. However, tooling for MBUnit is far from ideal. I have not been able to find any plugin that works with VS2015, neither for Resharper or pure runner (TestDriven.Net doesn't seem to work anymore). I don't feel very comfortable not being able to run tests. 

1) Are there any MBUnit runners for VIsual Studio 2015? Or command-line/gui?
2) Would it be an idea to consider moving to a more actively developed test framework, e.g. NUnit, XUnit, MSTest, etc?

I really love RoundhousE, and use it in production with success. I just fear depending on dying(?) frameworks could be a hindrance going forward. Opinions or tips?

ferventcoder:
We were looking at one point at converting all of the tests over to NUnit/TinySpec framework. If you are interested, look at DropKick or Chocolatey's tests and you can see how it is somewhat similar with less magic involved. 

You can run the tests from the command line by typing `test` and `test all`

erikbra:
If I should have any hope on working a bit on RoundhousE, I need a working test suite that can be run in Visual Studio. Do you know, @ferventcoder, if there are any active efforts to convert to a more actively developed testing framework, or should I just have a go at it? I have experience from day-job with xUnit and e.g. FluentAssertions, and no experience with TinySpec. Do you have a strong opinion on this? I just feel that if more "main-line" test and assert frameworks are not adopted, it will be difficult to get new contributors to engage. But, I might of course be biased from my current experience.

Do you have any strong opinions against converting to e.g. xUnit or NUnit? I havent looked at all the tests, and I think MBUnit is way more BDD-style than these two, but with the help of e.g. FluentAssertions, maybe we could make the porting not too intrusive.

ferventcoder:
@erikbra Chocolatey's test suite has that ability - and also the documentation is getting more comprehensive, see https://github.com/chocolatey/choco/blob/master/CONTRIBUTING.md#debugging--testing

erikbra:
I made a pull request where I converted all the MBUnit tests to xUnit. Could you please have a look at it when you have some time?

"Converted all tests from MBUnit to xUnit" #271

ferventcoder:
NUnit was the hope here.

erikbra:
[...]

PR #271, opened 2017-06-13
Issue #243: Converted all tests from MBUnit to xUnit

erikbra:
Issue #243: Converted all tests from MBUnit to xUnit (w/FluentAssertions and NSubstitute)

I converted all MBUnit tests to xUnit tests (I think it is a better fit than NUnit, as the `[Fact]` attribute to me is semantically similar to `[Observation]`. I'm not a hard-core BDD-er, so I'm open to that I might have misunderstood things.

I couldn't run `build.bat` or `test.bat` after the conversion, but I'm not very fluent in UppercuT, so I might need some help tweaking the build files. Everything builds fine in Visual Studio, and using MSBuild on the `RoundhousE.sln` file (same version used in `build.bat`.

Please take a look, and tell me what you think. I don't think the conversion involved too many changes, and I think using a test framework that is a) actively maintainted and b) known to most .NET developers is very important.


ferventcoder:
I was hoping for NUnit, not xUnit. But I will let @BiggerNoise be the decider of this.

ferventcoder:
FWIW, please update the uppercut.config to let it know of the new testing framework, otherwise the build will start failing.

erikbra:
Sorry about the xUnit/Nunit disappointment. I don't think it's too big a change to migrate to NUnit if it's important to you. I didn't find a good fit replacement for the `Observation`  attribute in NUnit (the observations are "statements" in naming), that's why I went for xUnit, but I'm open to discussion, of course :)

ferventcoder:
Minor nitpick - Git has commit summaries and bodies. The summary should be the first line, and around 50 characters. Then skip a line and write a good commit body, wrapping at 72 chars.

erikbra:
Love nitpicking. I will fix. Do you want me to squash the commits into one as well when I have to force push anyway? Or would you like the "Updated Nuget.exe" one to be a separate one, to make history explicit?

ferventcoder:
The preference for me (not sure about others) is that each commit should represent what it is about. If changing nuget.exe has nothing to do with other things you are doing, those are separate commits. 

Squashing comes when you are fixing smaller things and making commits, but all of those commits are really about the same thing (all considered part of an atomic change). "Like fixed compilation errors due to warnings" (acdd64d) is a great candidate for squashing.

BiggerNoise:
I'm assuming that this P/R is obsolete since you created the Nunit version. If I'm wrong on that, please reopen.

erikbra:
Hi, yes, this one is obsolete (or, if you wish, an alternative), since you have merged the NUnit one. Great to have some merges again! :) Closing this one is just fine.

PR #272, opened 2017-06-14
Consolidate DryRun items

rgrstvcr:
Consolidate all scripts contents that would have run into a single file when using the DryRun option.

This would be useful for revision prior the actual script migration.

### DIFF
--- a/product/roundhouse/runners/RoundhouseMigrationRunner.cs
+++ b/product/roundhouse/runners/RoundhouseMigrationRunner.cs
@@ -1,6 +1,7 @@
 namespace roundhouse.runners
 {
     using System;
+    using System.IO;
     using databases;
     using folders;
     using infrastructure;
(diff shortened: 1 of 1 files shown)


BiggerNoise:
@ferventcoder - Is there a way to see what the automated check is complaining about?

BiggerNoise:
@rgrstvcr Thank you for for your contribution, sorry for the delay in responding.

Can you elaborate on the use case for this P/R?

Is the plan:

1. dryrun rh and get the consolidated script?
2. review the changes that rh is going to make, revising the source as needed
3. run rh w/o dryrun

Steps 1 & 2 would repeat until you were happy with the changes that would be applied. Is that how you would intend to use it?

BiggerNoise:
@rgrstvcr - I'm not very clear what the purpose of this P/R really is. I'm going to close it for now.

I am fine with re-opening it, but I need to better understand the problem that this is going to solve for you. Please reply here to continue the discussion.

Issue #122, opened 2013-09-23
bdddoc.bat Throws an exception during doc generation


kolupaev:
bdddoc.bat Throws an exception of type ReflectionTypeLoadException.
It seems that it is caused by incompatible target platrofm of bdddoc.console.exe utility.

The problem can be fixed by recompiling bdddoc.console.exe with x86 target.

The exception log is attached as a screenshot.
![image](https://f.cloud.github.com/assets/533414/1196109/9f1af9bc-24a4-11e3-8e62-f2275280c8ac.png)

BiggerNoise:
We're not using bdddoc any longer

PR #273, opened 2017-06-19
Issue 243 - nunit version

erikbra:
As requested in Pull Request #271 , I have converted the tests to NUnit instead of xUnit. I have used TinySpec, as in Chocolatey tests, and tried to change as little as possible. Goal was to easily run tests in a modern development environment (Visual Studio), without having to use a command-line test runner. 

Could you please take a look at this pull, @ferventcoder , and see if you like this one better than #271?


BiggerNoise:
I will look at pulling this in as soon as I am sure that I did not mess up the 0.8.7 release.

Thank you so much for the work on this, bear with me for a bit longer.

erikbra:
Lovely, @BiggerNoise. Thanks. I wouldn't want to mess up the 0.8.7 release - it's been in the making for years :)
I say don't stress the merging before the release. There's nothing stopping release 0.8.8 coming in not too long :) 

The build failed because I had used C# 6 features. Any reason we want to stay on C# 5? We can handle this in a separate issue, anyway.

erikbra:
One thing, though. I might have broken some bdd doc generation? I don't know how to test it. But I don't think it has been working for ages, ref. #122.

ferventcoder:
Bdddoc was built to parse xml output from mbunit - so it would be expected to be broken when you switch out the testing framework (and that is okay).

PR #282, opened 2017-10-13
Remove dependency on machine installed SQL Server Compact

ErikEJ:


ErikEJ:
Build fail 😲

BiggerNoise:
@ferventcoder - Can we get the uppercut build log included into the TC artifacts? The TC build log is pretty useless.

BiggerNoise:
@ErikEJ Thank you for submitting this, I will try and merge this over the weekend. 

Briefly looking, It looks like you were able to use the official Microsoft nuget package. This should alleviate the concerns that @erikbra raised, correct? Or is there something else that I need to consider?

ErikEJ:
Correct, using latest version of official Microsoft nuget package

BiggerNoise:
Awesome, thank you for this change.

I'll try and figure out what's going on with the Team City build and get this merged over the weekend.

ErikEJ:
Thanks - Happy to help if I could get access to the build log!

erikbra:
Yes, sorry, guys. I didn't realize Sql Server CE and Compact Edition were the same. Official Microsoft Nuget packages are superduper. Lovely to avoid the dependency on an install.

BiggerNoise:
@ferventcoder - My JetBrains account is the same biggernoise if you want to just give me access to the project.

BiggerNoise:
OK. This seems to reproduce every time I run locally, so we don't need to crack the Team City Vault.

If you run build.bat or build.sh from the command line, it will fail and give you a build.log in the build_output directory.

The problem seems to be the post build steps that are trying to copy the CE libraries. Those fail when the build is run from the command line. I am not sure how to fix this or if we even need to supply those DLLs. (It's been a long week and my brain is a bit fuzzy).

At any rate, that should give you something to sleuth upon. Good luck!

ErikEJ:
Why does build.bat require .NET 3.5? I do not have that on my Windows 10 machine !?

BiggerNoise:
I would guess that dependency comes from NAnt as the foundation of the build system. We'll be looking into that shortly.

PR #283, opened 2017-10-16
Changed "fileencoding" parameter to "defaultencoding" and its behavio…

Liwoj:
…r to detect BOM 1st before using encoding from configuration

- default encoding (if not provided in configuration) changed from Encoding.Default (ansi code page configured in OS) to UTF-8
- fixed incorrect BOM detection of LE\BE variants of UTF-16 and UTF-32


BiggerNoise:
Thanks for getting this together!

Liwoj:
Pleasure :)

btw [wiki page](https://github.com/chucknorris/roundhouse/wiki/Scriptsfileencoding) needs to be updated too + some note in release notes about potential breaking change for people using ANSI encoding would be great.
I can update wiki too. Do you prefer to do it directly or through wiki repo ?

BiggerNoise:
If you can leave me some verbiage for the release notes, I would be glad to incorporate it.

For the wiki, I think we can start with direct updates. We may want to get a more formal process going in the future; for now, I'm taking whatever I can get for documentation purposes.

Issue #284, opened 2017-10-17
Revise Build Process to no longer require NAnt or .NET 3.5


BiggerNoise:
Current build is built upon UppercuT and thus NAnt. NAnt demands .NET 3.5 and has become more out of date than is reasonable to continue using.

The purpose of this issue is gain consensus on where we want to go next. @ferventcoder has mentioned wanting to migrate to [Cake Build](https://cakebuild.net/).

Personally, I am skeptical of the need for build frameworks. I prefer a plain old make/msbuild file.

Let's throw some arguments around.

ferventcoder:
UppercuT is what this should use, since both are part of the chucknorris framework - my proposal is that uc moves to cake build.

ferventcoder:
NAnt can be ran under .NET 4.x, one only needs to upgrade uc

erikbra:
I personally agree with @BiggerNoise . I don't see what value UppercuT adds to the build process, in addition to MSBuild. MSBuild is programming in XML, which is a nightmare. But, MSBuild is actively developed, and recently released as Open Source, and for all platforms, which I think is an advantage, should we move towards .NET Core and multi-platform support (which I think would be lovely). Even SQL Server is available on Linux now.

The other sweet thing about MSBuild, is it's the same build format that Visual Studio uses, and then you can have full parity between your IDE and command-line builds.

That said, I am very open to that I don't see all the facets of this, and UppercuT does look like a beast that can do just about everything. I'm just a bit skeptical to the NAnt foundation, as that is a bit old, and even Ant, which it is based upon, is widely replaced with other tools (e.g. maven) in the Java community too.

I think it would be fruitful to do a quick sketch of the build process of RoundhousE. Which steps are involved, and what are its deliverables. Then we can see which tools can help us performs these steps in the simplest fashion.

A requirement of a new build tool/build pipeline, IMHO, would be better integration with TeamCity, i.e. that the build logs are easily available in the TeamCity UI, and preferably with good step-by-step integration with TeamCity (can be achieved with [Service Messages](https://confluence.jetbrains.com/display/TCD10/Build+Script+Interaction+with+TeamCity) in just about any tool.

It would also be benefitial if the tools are widely installed, as it would make it easier for new developers to join the project. There is something elegantly simple with just cloning a repo and opening it in VS, being able to build and run tests at once.

Do you have any previous experience with Cake, @ferventcoder ? Are you using it in other projects, and have good experience with it?

[...]

Issue #274, opened 2017-07-16
Failing Chocolatey Moderation


BiggerNoise:
I'm trying to clear some of the old backlog. 0.8.7 has been released, but is failing moderation on the Chocolatey site. Specific issues are:

Please remove these files from your package:

content\ReadMe.RoundhousE.txt
docs\GettingStarted.docx
docs\LICENSE
docs\NOTICE

Because you have included binaries, you need to include tools\LICENSE.txt and tools\VERIFICATION.txt (see choco new for examples)

I have assigned @ferventcoder, but anyone can make the requested changes.

I'm going to pull in some of the outstanding pull requests to make an 0.8.8 and I'd like to release this to choclatey at the same time.

ferventcoder:
That's interesting but I get why those things would be pulled out.

erikbra:
Are we OK with calling this "done" for now, @BiggerNoise, @ferventcoder ? I think what is output from AppVeyor now in master should be approved in Chocolatey? I'm marking this as 0.9.0, supposing that you're OK with closing it. Please scream out if you disagree :)

BiggerNoise:
It should close with the 0.9.0 release. I want to hold off closing it for real until we actually get through the moderation process.

BiggerNoise:
The 0.9.0 release has passed the automated chocolatey automation. I'll close this once I can install 0.9.0 from choco

ferventcoder:
Don't forget to edit that nuspec to not just have me as the software author. It's more folks, but you both definitely should be listed in there.

ferventcoder:
For next time.

BiggerNoise:
['nuff said](https://chocolatey.org/packages/roundhouse/0.9.0)

PR #285, opened 2017-10-20
Convert to sdk-style projects

erikbra:
Microsoft introduced a new form of .csproj files related to the introduction of .NET core. These new, simplified files can also be used for "classic" .NET projects. The .csproj format is much simpler, files are globbed by default (no need to name each file of the project in the .csproj), and it has support for having nuget references directly in the .csproj file as `<PackageReference>` elements, instead of using packages.config.

This is the csproj file format going forward, and the MS tools will be optimized for this format.

I did have a couple of issues, though. The current build process had a few dependencies to old ways of doing stuff.

1) nuget restore is not called explicitly before build. This is the new/preferred way of doing it
2) NAnt tasks are hardwired against an old version of msbuild - we need a new one to be able to understand the new csproj format 
3) I couldn't get the build system to work properly for the output folders. I changed the output folders so that you can just build with `msbuild /p:Configuration=Build` and everything works, but the UppercuT things I couldn't get to work, I need some help.
4) I do think that UppercuT (or possible it's the Nant coupling) is extremely tiresome to work with. We have a build process consisting of 4-5 steps, and modifying the templates in XML is really a drag. I created an alternative `build_new.ps1`, which does everything up to and including ILMerge, and I produce a running rh.exe. But I don't expect you to accept the pull request as is, as the standard `build.bat` doesn't work properly. I would love if @ferventcoder could take a look at what's wrong. 

I would appreciate some feedback from both @BiggerNoise and @ferventcoder on this PR. Is this an OK size for a PR related to #284, or is it too big? The problem is, Nuget package versions are tightly coupled to Nuget version, which is coupled to MSBuild version, which is coupled to the .csproj file format. So is's difficult to change one of them in isolation.


erikbra:
I forgot to mention, I had a problem with strong-naming. The latest versions of Nuget dependencies (i.e. StructureMap) does not come strong named (signed), and will not be releasing any more strong named versions. Strong naming had its place in time, but I don't think it's relevant anymore. Do any of you think it's important to strong-name rh.exe? If we do, there's a nuget package that can be installed in all projects using non-signed nuget packages, which signes them on-the-fly [https://github.com/dsplaisted/strongnamer](https://github.com/dsplaisted/strongnamer). We can consider using that if you have strong arguments for strong name signing rh.exe.

BiggerNoise:
@erikbra - Thank you very much for this. As a rule, I would want P/Rs to be smaller, but, as you said, there's no real good way to break this up. I didn't see anything that looked out of place in the changes.

I had forgotten that we even did strong naming; I'm fine with losing it.

@ferventcoder - I think we need to attach a fair amount of urgency to this P/R. I think the recent burst of activity in RH might motivate others to create P/Rs. Merging P/Rs and this is going to be an epic pain.

Assuming that I can't talk you out of cake based builds, I would propose the following: 

 - If you can get the cake going this weekend, add that to this P/R and merge it. 
- Otherwise, let's merge this as is and remove the old uppercut build (use the powershell file to build). We'll add an "Under Construction" blurb to the read me and you can take a couple of weeks to get the real build system in place. That will allow P/Rs to proceed that are unlikely to have massive merge conflicts.

Sound like a plan?

erikbra:
Thanks a lot, @BiggerNoise . I do understand the general rule of thumb. But it gets complicated when packages and tools and build systems all depend on each other. I see that I have left out a bit in the powershell based build, I stopped at creating rh.exe. The nugets are never built. But, if you're fine with having powershell based builds for now (an intermediary), I will complete it so that we get an actual deployable out in the other end.

What concerns me, is that we have no build server that can build this, it fails on the one we're using at teamcity.org, because of too old nuget version: 

> The 'Newtonsoft.Json 10.0.3' package requires NuGet client version '2.12' or above, but the current NuGet version is '2.8.60717.93'.
> The 'System.Reflection.Emit.Lightweight 4.3.0' package requires NuGet client version '2.12' or above, but the current NuGet version is '2.8.60717.93'.

Who maintains the build server, @ferventcoder? Is it you? Or is it Jetbrains, in some way?

BiggerNoise:
I have nothing to do with the Team City's stuff, so it is undoubtedly @ferventcoder. 

I'd really like to merge this in and update the README to indicate that it is under construction. 

@ferventcoder  - If you want to add me to the Team Cities account, I can try updating things there. Or if you think you'll have the preferred build up and running shortly that would be great also. I just don't want this P/R to linger.

erikbra:
I have now fixed the `build.ps1` script so that it both adds proper SemVer (using GitVersion.exe - maybe we need to resolve how to install that - should we assume it's installed already via e.g. chocolatey, or should we ship binary in the repo?), and it also generates a .nupkg.

I have chosen a bit different approach to building the nupkg, as msbuild now supports this natively. I added project dependencies to all `roundhouse.<databasespecific>.csproj` from `roundhouse.console.csproj`. I don't see any disadvantages of doing this. 

Then I added the ILMerge task as part of the `roundhouse.console.csproj` file. Personally I think this is a *lot* less complex than using UppercuT, but I don't want to run over you, @ferventcoder, as this is your project. Just, please have a look at how simple (imho of course, but I created it, and might be a bit blind) the build.ps1 file is. It just calls gitversion, msbuild to build and pack, and nunit-console to run tests. No more xml config. Just give it a glance before you go all in on the Cake track.

MSBuild has matured significantly since MS took it in, and I think using the same build script for Visual Studio and cmd line has great advantages. But YMMV.

This won't build on the TeamCity build agent we have, as it requires the newest Visual Studio toolchain (Visual Studio 2017 and Nuget 4.3+).

BiggerNoise:
Thanks again for all your work on this. Mondays are really busy for me, so I'll do some roundhousing tomorrow night.

ferventcoder:
> Otherwise, let's merge this as is and remove the old uppercut build (use the powershell file to build). We'll add an "Under Construction" blurb to the read me and you can take a couple of weeks to get the real build system in place. That will allow P/Rs to proceed that are unlikely to have massive merge conflicts.

Sounds good to me.

ferventcoder:
> Who maintains the build server, @ferventcoder? Is it you? Or is it Jetbrains, in some way?

Jetbrains. But it's as simple as updating the nuget in the lib folder to a newer version.

ferventcoder:
Run with what you got for now. Cake is probably at least a few weeks out.

You are right, the current UC stuff is a pain to work with and it doesn't work with newer infrastructures due to a lack of updated NAnt. And it is unlikely to ever be updated given most folks hate the pain of xml.

Don't forget all the things that go into the packaging parts, that's where much important work is done in physically separating build output from what goes in packages.

nuget/roundhouse/roundhouse.nuspec
```diff
@@ -12,6 +12,15 @@
     <licenseUrl>http://www.apache.org/licenses/LICENSE-2.0</licenseUrl>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
     <tags>roundhouse db migration database migrator chucknorris</tags>
-	<iconUrl>https://raw.github.com/chucknorris/roundhouse/master/nuget/RoundhousE_Logo.NuGet.jpg</iconUrl>
+    <dependencies />
+	  <iconUrl>https://raw.github.com/chucknorris/roundhouse/master/nuget/RoundhousE_Logo.NuGet.jpg</iconUrl>
   </metadata>
+  <files>
+    <file src="$mergedExe$" target="bin\" />
+    <file src="$appConfigFile$" target="bin\" />
```
ferventcoder: Ouch, we need a config file now?

nuget/roundhouse/roundhouse.nuspec
```diff
@@ -12,6 +12,15 @@
...
     <tags>roundhouse db migration database migrator chucknorris</tags>
-	<iconUrl>https://raw.github.com/chucknorris/roundhouse/master/nuget/RoundhousE_Logo.NuGet.jpg</iconUrl>
+    <dependencies />
+	  <iconUrl>https://raw.github.com/chucknorris/roundhouse/master/nuget/RoundhousE_Logo.NuGet.jpg</iconUrl>
   </metadata>
+  <files>
+    <file src="$mergedExe$" target="bin\" />
+    <file src="$appConfigFile$" target="bin\" />
+    <file src="..\..\nuget\roundhouse\content\ReadMe.RoundhousE.txt" target="content\" />
+    <file src="..\..\docs\GettingStarted.docx" target="docs\" />
```
ferventcoder: What do you think about leaving this one out?

nuget/roundhouse/roundhouse.nuspec
```diff
@@ -12,6 +12,15 @@
...
+    <file src="..\..\docs\legal\*" target="docs\" />
```
ferventcoder: Might be good to move this to next to the binaries, and move the binaries from the bin file?

nuget/roundhouse/roundhouse.nuspec
```diff
@@ -12,6 +12,15 @@
     <licenseUrl>http://www.apache.org/licenses/LICENSE-2.0</licenseUrl>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
     <tags>roundhouse db migration database migrator chucknorris</tags>
-	<iconUrl>https://raw.github.com/chucknorris/roundhouse/master/nuget/RoundhousE_Logo.NuGet.jpg</iconUrl>
+    <dependencies />
+	  <iconUrl>https://raw.github.com/chucknorris/roundhouse/master/nuget/RoundhousE_Logo.NuGet.jpg</iconUrl>
   </metadata>
+  <files>
+    <file src="$mergedExe$" target="bin\" />
```
ferventcoder: What do you think of moving this to the tools directory or a subdirectory of that?

product/roundhouse.console/Properties/AssemblyInfo.cs
```diff
@@ -4,4 +4,4 @@
 // The following GUID is for the ID of the typelib if this project is exposed to COM
 
 [assembly: Guid("a7a8d35d-cb38-4f12-a388-41c0fd21bf20")]
-[assembly: AssemblyKeyFile("..\\..\\RoundhousE.snk")]
\ No newline at end of file
+//[assembly: AssemblyKeyFile("..\\..\\RoundhousE.snk")]
```
ferventcoder: Signing is less important nowadays to most folks in the community. The question is whether organizations (the quiet unheard group) still feel the same way. I like the concept of that signer you mentioned.

product/roundhouse.console/roundhouse.console.csproj
```diff
@@ -1,65 +1,44 @@
...
     <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\..\</SolutionDir>
-    <RestorePackages>true</RestorePackages>
-    <TargetFrameworkProfile />
+    <NoWarn>NU1701</NoWarn>
+    <GenerateBindingRedirectsOutputType>true</GenerateBindingRedirectsOutputType>
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+    <CodeAnalysisRuleSet>AllRules.ruleset</CodeAnalysisRuleSet>
+
+    <VersionPrefix>0.8.9</VersionPrefix>
+    <VersionSuffix>alpha</VersionSuffix>
```
ferventcoder: How do these get updated?

product/roundhouse.console/roundhouse.console.csproj
```diff
@@ -68,115 +47,93 @@
...
+     <ILMergeDir>..\..\code_drop\merge\</ILMergeDir>
+     <ILMergeTarget>$(ILMergeDir)$(AssemblyName).exe</ILMergeTarget>
+  </PropertyGroup>
+
+  <Target Name="ILMerge" BeforeTargets="Pack">
+    <ItemGroup>
+      <ILMergeAssemblies Include="$(OutputPath)*.dll" />
+    </ItemGroup>
+    <PropertyGroup>
+      <IgnoreFile>..\..\build.custom\ilmerge.internalize.ignore.txt</IgnoreFile>
```
ferventcoder: Consider moving this somewhere more appropriate over time.

product/roundhouse.console/roundhouse.console.csproj
```diff
@@ -68,115 +47,93 @@
...
+      <ILMerge>..\..\lib\ILMerge\ILMerge.exe</ILMerge>
```
ferventcoder: Where does this resource come from?

product/roundhouse.console/targets/Microsoft.Application.targets
```diff
@@ -226,11 +219,11 @@ Copyright (C) 2005 Microsoft Corporation. All rights reserved.
   <Target Name ="_BuiltExeOutputGroupOutput"
           Condition="!$(Disable_CopyExeApplication)"
           Outputs="@(BuiltExeOutputGroupOutput)">
-    <CreateItem Include="$(ExeProjectOutputDir)\bin\**\*.*;@(Content->'%(FullPath)')" Condition="'$(OutDir)' == '$(OutputPath)'">
+    <CreateItem Include="$(ExeProjectOutputDir)\bin\**\*;@(Content->'%(FullPath)')" Condition="'$(OutDir)' == '$(OutputPath)'">
       <Output ItemName="BuiltExeOutputGroupOutput" TaskParameter="Include"/>
     </CreateItem>
 
-    <CreateItem Include="$(ExeProjectOutputDir)\**\*.*" Condition="'$(OutDir)' != '$(OutputPath)'">
+    <CreateItem Include="$(ExeProjectOutputDir)\**\*" Condition="'$(OutDir)' != '$(OutputPath)'">
```
ferventcoder: 👍  thanks for fixing these

product/roundhouse.databases.sqlite/roundhouse.databases.sqlite.csproj
```diff
@@ -1,156 +1,87 @@
...
+        <None Include="App.config"/>
+    </ItemGroup>
+    <ItemGroup>
+        <EmbeddedResource Include="$(NuGetPackageRoot)System.Data.SQLite\1.0.84\lib\net45\System.Data.SQLite.dll">
+            <Link>assemblyresource\System.Data.SQLite.dll</Link>
+        </EmbeddedResource>
+    </ItemGroup>
+    <Import Project="$(NuGetPackageRoot)System.Data.SQLite.Core\1.0.105.2\build\net45\System.Data.SQLite.Core.targets"
+            Condition="Exists('$(NuGetPackageRoot)System.Data.SQLite.Core\1.0.105.2\build\net45\System.Data.SQLite.Core.targets')"/>
 </Project>
```
ferventcoder: Are the SQLite Interops not necessary anymore? They were native DLLs and unable to be merged, so they were embedded and unpacked at runtime.

BiggerNoise:
On the single DLL, output location, the interops, and version. I would defer worrying about those to when we implement the real build solution. 

I'm open to leaving the docs out of nuget.

I'll open an issue for the signing. You're right that there are probably people more corporate than us that might really care, let's see if anyone speaks up.

For now, I just want to be able to build and test with the csproj files. I'm going to bring this in tonight and create issues for the shortcomings that we want to address. Thanks for getting back to us on this one, good to know that I'm not completely off in the weeds.

nuget/roundhouse/roundhouse.nuspec
erikbra: Maybe not? Looks like there is mostly bollocks in it. It was in rh 0.8.8 (the latest published one). I can leave it out, if we don't want it.

nuget/roundhouse/roundhouse.nuspec
erikbra: By all means. I saw there was an issue that you should have left out some files to be Chocolatey "compatible" anyway. (#274).  I'll take out all but the ones mentioned in that one.

nuget/roundhouse/roundhouse.nuspec
erikbra: Which "this one" do you mean? The legal\* ones?

product/roundhouse.console/roundhouse.console.csproj
erikbra: They don't. I override them on command-line with MSBuild properties when packing (in build.ps1). I generate semantic version using GitVersion.exe (probably same logic as in the nant build files) . Maybe just leave them as a "default" or easy-to-understand-that-is-dummy value in the .csproj?

Do you have any thoughts on external dependencies, such as GitVersion.exe? Should we check in the binaries, or should we expect people to use chocolatey to install themselves?

product/roundhouse.console/Properties/AssemblyInfo.cs
erikbra: I can look into it. It basically signes all unsigned dlls in the project. But we'll need to use it in all the projects, then, bercause the DLL changes (it's basically a different one, as the Assembly qualified name changes).

nuget/roundhouse/roundhouse.nuspec
erikbra: I don't know what chocolatey expects. The nuget of the exe is basically for Choco, isn't it?

product/roundhouse.console/roundhouse.console.csproj
erikbra: Of course. We could move all these up to the roundhouse.console project (maybe in a "pack" or "nuget" subfolder).

product/roundhouse.console/roundhouse.console.csproj
erikbra: Which resource? ILMerge.exe?

product/roundhouse.databases.sqlite/roundhouse.databases.sqlite.csproj
erikbra: I really don't know, I might have missed something. That is a challeng, as I don't know if we have a working integration test environment for all DBs. Do we? If not, maybe we should set up something. People task well of containers. Maybe we could use them for this? (just an idea - we should create a separate issue for this)

erikbra:
Thanks for all the comments, @ferventcoder. I'll look into them and update. I could really need some help running integration tests on all DBs, as I don't have environments for all this. When talking buildserver, we really need to update more of the tools. Nuget needs to be 4.3+ (can be solved by updating binary), but we also need the VS2017.4 (15.4) build tools, as msbuild has lots of new functionality there. How do we control this? Can you set necessary "Capabilities", and then they dynamically spin up a VM, or is it a manual process?

ferventcoder:
> Nuget needs to be 4.3+ (can be solved by updating binary), but we also need the VS2017.4 (15.4) build tools, as msbuild has lots of new functionality there. How do we control this? Can you set necessary "Capabilities", and then they dynamically spin up a VM, or is it a manual process?

In AppVeyor, you can look at what they have, and then just choco install what is not there. https://www.appveyor.com/docs/build-environment/ (note there is a VS 2017 environment ready to go https://www.appveyor.com/docs/build-environment/#build-worker-images). All the preinstalled software - https://www.appveyor.com/docs/build-environment/#pre-installed-software.

Looks like NuGet 4.3 is [preinstalled as well](https://www.appveyor.com/docs/build-environment/#tools): 

image

ferventcoder:
I know I've been telling you AppVeyor is awesome, and now @jbogard is as well. Trust us, it's awesome. 👍

nuget/roundhouse/roundhouse.nuspec
ferventcoder: Meaning instead of the docs folder as the location in the nupkg, the license and notice and other legal ones should be beside the binaries.

nuget/roundhouse/roundhouse.nuspec
ferventcoder: It doesn't matter where it is, but what the consistent approach has been for many packages is a subfolder of tools.

product/roundhouse.console/Properties/AssemblyInfo.cs
ferventcoder: 👍

product/roundhouse.console/roundhouse.console.csproj
ferventcoder: GitVersion is awesome.

ferventcoder: There is a concept of setup, where all the tools are brought to a repository. That setup should give a developer everything they need to do build.

product/roundhouse.console/roundhouse.console.csproj
ferventcoder: More of a placeholder comment... just thinking about where they could go, no action to be done now.

product/roundhouse.console/roundhouse.console.csproj
ferventcoder: `<ILMerge>` - typically if you use things from MSBuild that are not included, it can cause failures. It's good to know if this is build into MsBuild or requires a community pack or something.

product/roundhouse.databases.sqlite/roundhouse.databases.sqlite.csproj
ferventcoder: We should set something, that would be good. AppVeyor would give us this integration testing for cheap.

nuget/roundhouse/roundhouse.nuspec
erikbra: That's an idea, but according to #274, they want the Licence one as LICENCE.txt in the tools folder (together with a VERIFICATION.txt). I see that the NOTICE file is just boiler-plate, so we don't need that, at least.

nuget/roundhouse/roundhouse.nuspec
erikbra: So the `rh.exe` itself should be in a subfolder of `tools`? and then it's the job of `chocolateyInstall.ps1` to do what it wants with it (copy to Program Files or AppData or whatever)?

ferventcoder: I'm not sure that chocolateyInstall.ps1 is even necessary.

ferventcoder: The `install.ps1` needs to go away.

ferventcoder: With just binaries in the package (no scripts), you can run these commands in succession:

* `choco install roundhouse`
* `rh`

Chocolatey automatically shims executables found in packages and puts them into a folder that is on the PATH (`$env:ChocolateyInstall\bin`).

product/roundhouse.databases.sqlite/roundhouse.databases.sqlite.csproj
ferventcoder: If we are not sure, best to leave those interop dlls there and let them embed in the binary.

nuget/roundhouse/roundhouse.nuspec
erikbra: Nice. I now know more why i love choco ;)
So, skip the chocolateyInstall.ps1. Where do we put rh.exe? Tools? root?

nuget/roundhouse/roundhouse.nuspec
ferventcoder: LICENSE can be brought in as is. It doesn't actually need the .txt, but it's helpful so that it can be read as text on Chocolatey.org package page under the files section.

nuget/roundhouse/roundhouse.nuspec
ferventcoder: My vote for consistency is the tools folder.

product/roundhouse.console/roundhouse.console.csproj
erikbra: True, true. We could make a `setup.ps1` which pulls chocolatey from the web, etc, and uses that to install GitVersion. but maybe a bit intrusive. Is it possible to install choco packages without admin? If not, this will definitely not work for build servers, etc. I love having the same script on build servers as local, so that would be nice, to create a setup.ps1, which assures all necessary tools are in place, and then the build.ps1 (or cake, maybe, preferably), could do the build part itself. 

nuget it part of the build chain, but the .exe of chocolatey is not in the nuget package (which makes sense).

product/roundhouse.console/roundhouse.console.csproj
erikbra: No, no, <ILMerge> is not a task. It is just a property. Maybe I should have called it <ILMergeExe> or something.

product/roundhouse.databases.sqlite/roundhouse.databases.sqlite.csproj
erikbra: I like cheap (at least on my spare time. Daytime I like paying for good tools)

erikbra:
Would you like me to initiate the AppVeyor journey, @ferventcoder ? Or, if you already use it, you could create an official "chucknorris" or "roundhouse" account, and give me project edit rights, then I could look at the tools needed to build in the big, mighty cloud.

nuget/roundhouse/roundhouse.nuspec
erikbra: So we end up with just this in the Nugget (I love calling them nuggets), then?

image

product/roundhouse.console/roundhouse.console.csproj
ferventcoder: The good news? Cake does all of this already.

product/roundhouse.console/roundhouse.console.csproj
ferventcoder: I see it now... property group. My bad...

erikbra:
@ferventcoder My email is erik@brandstadmoen.net

ferventcoder:
Redacted the email address, you need an account with AppVeyor, but I think you should be good just to go to https://ci.appveyor.com/login

image

ferventcoder:
Fixed image

erikbra:
I have a public email address, no worries, but thanks anyway. I'll sign in and see what I get.

BiggerNoise:
@ferventcoder - I have created an AppVeyor account and sent you the email that I prefer.

@erikbra - I'm having mixed results when I pull this down. The build is fine in visual studio and the command line, but it looks like the console pack and unit test aren't going correctly. This is what I am seeing here:
```
C:\Program Files\dotnet\sdk\2.0.0\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(141,5): error : Could not find a part of the path 'C:\development\Chuck\roundhouse\code_drop\merge'. [C:\development\Chuck\roundhouse\product\roundhouse.console\roundhouse.console.csproj]

 * Looking for nunit3-console.exe
    - Found at C:\Users\bigge\.nuget\packages\nunit.consolerunner\3.7.0\tools\nunit3-console.exe

 * Running unit tests

dir : Cannot find path 'C:\development\Chuck\roundhouse\build_output\RoundhousE.UnitTests\net461' because it does not exist.
At C:\development\Chuck\roundhouse\build.ps1:42 char:13
+ $tests =  $(dir "$($TESTOUTDIR)\*.tests.dll");
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (C:\development\...nitTests\net461:String) [Get-ChildItem], ItemNotFoundException
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.GetChildItemCommand

Runtime Environment
   OS Version: Microsoft Windows NT 10.0.16299.0
  CLR Version: 4.0.30319.42000

Error: no inputs specified
```

Unfortunately, I got a late start tonight and my brain is kind of mush. You would expect this to work, right? I installed GitVersion, and added msbuild to my path so I can hit it from powershell; was there something else I was supposed to do?

BiggerNoise:
I can see the project in AppVeyor, so gtg there.

erikbra:
@BiggerNoise Probably forgot to create some folders if not already existing, in the msbuild script. I'll do a git clean and try again.

erikbra:
Try again now, @BiggerNoise . There was some confusion btw Solution configurations (Which are "Build|Any CPU") and project configurations (which are "Release|AnyCPU"). There are no "Build" project configurations. "Build" solution config is set to build the "Release" project configs. In addition, there was the AnyCPU/Any CPU (with blank) confusion, as described here: http://jimmyscorner.com/archives/51/the-differences-between-solution-build-configurations-and-project-build-configurations

This made the folders not align up, as I thought the conditionals on Configuration = "Build|AnyCPU" would hit. They didn't. But now it should work from scratch.

BiggerNoise:
@erikbra I don't think that your commits made it to the P/R

erikbra:
Blaargh, sorry. I may have forgotten to push. I'll check when I get to that computer in the morning (CET).

erikbra:
Yepz, n00b mistake. Commit, move on to something else. Forgot to push. Now it should build from scratch, @BiggerNoise .

BiggerNoise:
I'm still having a couple of issues building from scratch, but don't have time to get into them this morning.

I updated the AppVeyor project to run .\build.ps1 (for now). 

It's reporting success, but it's lying. Let me know if you have trouble seeing the logs.

erikbra:
Yes, it should fail on error output. I don't know how to configure that in AppVeyor, only TeamCity. Seems like some of the folders still don't exist beforehand. I'll see if I can reproduce if nuking my ~/.nuget folder. Didn't think of that one. And code_drop\log\ is missing (some of it), apparently, too.

erikbra:
We have too old MSBuild in AppVeyor, @BiggerNoise. We need 15.0. See log from latest build: 

> C:\projects\roundhouse\product\roundhouse\roundhouse.csproj.metaproj : warning MSB4078: The project file "product\roundhouse\roundhouse.csproj" is not supported by MSBuild and cannot be built.

lahma:
@erikbra have tried defining [appveyor.yml](https://github.com/quartznet/quartznet/blob/master/appveyor.yml) with vs2017 image?

ferventcoder:
@lahma We will get the appveyor.yml file exported to this source once it is all set up. Currently it is defined as follows:

~~~yml
version: 1.0.{build}
image: Visual Studio 2017
build_script:
- ps: .\build.ps1
~~~

lahma:
@ferventcoder Now that I actually read the [output](https://ci.appveyor.com/project/chucknorris/roundhouse/build/1.0.6) from appveyor, it seems that MSBuild should be good to go.

`
MSBuild auto-detection: using msbuild version '15.4.8.50001' from 'C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\bin'.
`

However,
`$nunit = $(dir -r $env:HOME\.nuget\packages\nunit* -i nunit3-console.exe | Select-Object -last 1)`

Does not seem to resolve to correct dir:
`dir : Cannot find path 'C:\.nuget\packages' because it does not exist.`

What about:

`$(dir -r $env:UserProfile\.nuget\packages\nunit* -i nunit3-console.exe | Select-Object -last 1)`

The next errors seems to originate from the fact that databases cannot be contacted, have they been enabled in build configuration?

Appveyor also does some automatic test detection magic so might conflict with build.ps1...

erikbra:
I have fixed the Nuget folder problem (look for `$env:NUGET_PACKAGES`, if not found, use `~/.nuget` (works on all platforms)). I have also removed the running of tests manually with nunit-console on AppVeyor (check for `$env:AppVeyour`), since it seems to pick up the tests itself. You just need to tell it not to run the integration tests, as we don't have any integration tests environments set up yet.

erikbra:
We have a lot to go on with regards to test setup, artifacts gathering, etc, but at least we're building on AppVeyor now. I would prefer if we could put the AppVeyor config in source. Easier to maintain. But maybe we should make a separate issue for that, @ferventcoder . We also have some cleanup tasks, to assure all the nuget packages created are OK, etc, etc (I think they're not right now).

I'm at 00:30 CET now, so I'm retiring for the day. I'll let the rest of the world work :)

PR #293, opened 2017-10-29
Removed some warnings in the build

erikbra:
Didn't know what to do with the last ones, as they're obsolete APIs actually exposes via command-line parameters. Should we remove them from the command line options too? They've probably been deprecated for 4 years, but has the deprecation been communicated to end-users of `rh.exe`, or can we expect someone to rely on those? And if, should we care? Or just announce a breaking change?

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -14,7 +14,8 @@ $onAppVeyor = $("$($env:APPVEYOR)" -eq "True");
 pushd $root
 
 
-"`n * Generating version number"
+"`n"
+" * Generating version number"
 $gitVersion = (GitVersion | ConvertFrom-Json)
 
 If ($onAppVeyor) {
@@ -24,7 +25,7 @@ If ($onAppVeyor) {
     appveyor UpdateBuild -Version "$newVersion"
 }
 
-"`n * Restoring nuget packages"
+" * Restoring nuget packages"
 nuget restore -NonInteractive -Verbosity quiet
 
 # Create output and log dirs if they don't exist (don't know why this is necessary - works on my box...)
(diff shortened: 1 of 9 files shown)


BiggerNoise:
I think we can simply announce the breaking change for the recovery mode, it's a pretty straightforward migration to the supported option.

erikbra:
OK, I'll remove the command-line parameters too, so we get rid of all the warnings.

ferventcoder:
Why environment name? That is hugely important

ferventcoder:
I see - names? I don't remember introducing that and deprecating the singular

erikbra:
The EnvironmentName was obsolete, got warnings in code. But was exposed via command-line. So all builds had lots of warnings. Should we re-add it? There is an `EnvironmentNames` property. Maybe we shouldn't remove the singular? I just assumed it was obsolete since it was marked as such.

erikbra:
Should I re-introduce the `EnvironmentName` (singular) property on the command line, @ferventcoder?

ferventcoder:
I don't remember which one goes, but passing the environment through cli is paramount.

ferventcoder:
Commit history is typically great to read - if commits were well done the commit introducing the deprecations will tell you the reasoning

ferventcoder:
One of the biggest reasons for git commits with summaries and bodies is so you understand why the change was made years from now when you've forgotten why you did something (as is the case with me right now).

ferventcoder:
I would say back this out completely for now - no wonder I don't remember it, the deprecation was only introduced into master 19 days ago.

ferventcoder:
https://github.com/chucknorris/roundhouse/issues/235

ferventcoder:
Actually reviewed again, we are maybe good. Extensive testing is what this one does need, but keep around the deprecation awhile longer, or mark as breaking change and be sure it is highlighted i  release notes. Deprecations IMHO should last longer than one release...

BiggerNoise:
The P/R that added environment names did come with a fair bit of testing. On the command line, the plural was just a synonym for environment name.

I have a couple of other questions that I'll add to #295

erikbra:
I made a "fix-PR", #296. Sorry again for the too early merge.

Issue #276, opened 2017-10-05
.NET Core support?


jbogard:
Wanted to use the RH library in a .NET Core project but no dice :(

BiggerNoise:
I am curious if you want to use RH as a library or if you just wanted a .NET Core executable?

At any rate, my (rough) plan is:
- Merge as many P/Rs as I can and release 0.8.8
- Modernize the dependencies and tools for release 0.9.0 (that would be when I could look at .NET Core)

No timelines, but hopefully, the gap between 0.8.7 and 0.8.8 will be smaller than the one between 0.8.6 and 0.8.7.

jbogard:
As a library, to be able to execute as part of running a suite of tests.

Dunno about the .net core side but a next step would be a .NET CLI tool.

lahma:
There might be some work to do, here's the [analysis for .NET Core](https://icanhasdot.net/result?github=chucknorris~2Froundhouse). Looks like Fluent NHibernate and NHibernate are the ones that require most love at the moment.

jbogard:
What is RH using FNH/NH for?

NightOwl888:
> As a library, to be able to execute as part of running a suite of tests.

In that case, it might make sense to target [.NET Standard](https://docs.microsoft.com/en-us/dotnet/standard/net-standard) rather than .NET Core, which gives you better platform support. .NET Standard applies only to libraries, but makes them possible to share between .NET Framework and .NET Core (which ultimately cuts down on platforms you need to target).

jbogard:
Yep! Just did this myself with my Respawn and BulkWriter projects, targeting netstandard1.2 I believe.

BiggerNoise:
@lahma Thanks for that information
@jbogard It's used to create and populate the tables in the roundhouse schema (scripts_run, etc.)

The use of NH is so limited in the lib, that I would almost lean towards not using an ORM at all. It's been a while since I have been deep into that area of the code, but that seems like a great way to lop off a dependency.

I'll stress that I want to get some of the P/R's merged an get a 0.8.8 out before I start wrenching in this area. And life seems to have other plans for me apart from doing Roundhouse.

jbogard:
Is that how RH provides support for different DBs? Lets NH handle that?

BiggerNoise:
I don't think NH is that involved in running the user scripts (other than as a pass through to get the connection).

We definitely use it for managing the scripts_run and other tables in the roundhouse schema.

[...]

Issue #16, opened 2011-09-10
RoundhousE does not update its own Tables


BiggerNoise:
Bit ironic really :)

We started with an old version of RH on our servers. Now when we get errors, we get the barf that RH cannot insert into the errors table (message is truncated).

Looks like Message field in the ScriptRunErrors table was changed from nvarchar(255) to ntext a while back, but RH doesn't version its own tables so we get the log barf instead of the real error message.

I can take a look at this, unfortunately, it may lead to a state of infinite recursion.

PR #294, opened 2017-10-29
Porting to netstandard2.0/netcoreapp2.0

erikbra:
Can we find a workaround?

There's a missing feature in netstandard2.0, the `DbProviderFactories.GetFactory(string)`. This one is used in `AdoNetDatabase.cs`. Can we work around this? There seems to be very much discussion and non-code-related discussions in the CoreFX repo regarding this, see e.g. 

https://github.com/dotnet/corefx/pull/19885

seems like fixing is pushed to netstandard2.1 (not yet announced).

Can we do something to work around this, so that we can get closer to solving #276 ?


jbogard:
You can ping @FransBuoma, he's probably hit this I think.

Another option is to target netcoreapp2.0 perhaps? Is there a TFN where this is available?

FransBouma:
The api is currently under review. I already wrote the code for .NET core, but they first wanted to create and debate an API. The first version they came up with was completely different from what I implemented in my implementation. API review: https://github.com/dotnet/corefx/issues/20903. PR: https://github.com/dotnet/corefx/pull/19885
The current API proposal is OK by me with some small adjustments, but it takes ages for them to decide these trivial things. No-one else has participated in the discussions as well, it seems there aren't a lot of people who still need this API it seems. (which isn't a surprise, considering everyone who ported their ORM to .netstandard 2.0 has had to implement a DbProviderFactory registration system one way or the other by now)

erikbra:
Thank you, @FransBouma, for commenting. I have read your issues and pull requests in the CoreFX repo, and I felt sorry for you. It seems like they are stuck in analysis paralysis. I am very aware that changing an API used by thousands of users is different than one used by only a few. Of course the quality and API design is important, but it must be frustrating trying to contribute when you perform the work, and then it just gets put off, waiting for reviews.

A couple of questions: 

1) Is there anything we as developers-using-the-api-and-wanting it back can do to support you pushing the change through?
2) Do you think it would be possible to release the PR as a Nuget Library with extension methods waiting for the merge, or is it too intertwined with Core code for this to be practical?

FransBouma:
@erikbra :) Yes it was very frustrating, also because no-one else outside MS participated. At the end it felt like "why am I even investing my precious free time in this when they make it unpleasant?" so I gave up on that. Their loss. Anyway, IF they ever finalize the api review it's easy to change the implementation a bit to meet it so the most work is done already. 

To answer your questions:

1. The best you can do is perhaps review the API proposal as it is now (so the one at the bottom) and if you have anything to say, even if it's OK by you, then state that or upvote things so they know outsiders agree with that statement/proposal. 
2. I wouldn't do that, as you then have something which likely won't match what's going to be in the API in the future. If you need DbProviderFactory today, you should simply create a method and a static ConcurrentDictionary<string, DbProviderFactory> today so your users first register the factory there with the method and your code simply pulls the factory from that dictionary on .NETstandard. That will work today on .netstandard 2.0. See for an example what we added to our orm for that:
https://www.llblgen.com/Documentation/5.3/LLBLGen%20Pro%20RTF/Using%20the%20generated%20code/gencode_runtimeconfiguration.htm#dbproviderfactory

That simply registers the type and implies the invariant name from the namespace (as that's what DbProviderFactories does today on .net full anyway) Internally I then obtain the factory from the concurrentdictionary and if it's not there on .net full try to read it from the DbProviderFactories class with GetFactory().

erikbra:
Thanks for your explanation, @FransBouma. I will definitely look into having a dictionary of string to `DbProviderFactory` implementations. Do you know it there is an authorative list on what string should map to what? Maybe the best place to look is my Windows registry? But where?

erikbra:
I have checked in a proposed workaround to the missing `DbProviderFactories.GetProvider`, as we always subclass the class using the factory, we might as well delegate the creation of the the factories to the subclasses. Wny thoughts, @ferventcoder or @BiggerNoise?

@jbogard, would you mind pre-alpha-testing if the netstandard support of roundhouse.lib would suit your needs? The built artifact is here: https://ci.appveyor.com/project/chucknorris/roundhouse/build/0.8.9-PullRequest.294+32.26/artifacts

(Also available on the following feed, but we're not advertising that a lot yet: https://ci.appveyor.com/nuget/roundhouse  package roundhouse.lib, version roundhouse 0.8.9-PullRequest.294)

FransBouma:
@erikbra 
>  Do you know it there is an authorative list on what string should map to what? Maybe the best place to look is my Windows registry? But where?

That's not needed :) The user has to register the type in all cases, as there's no central registration system on netstandard/netcore, so the only source for this information, what type of DbProviderFactory to register and with which name, is up to the user. The user has to register a type as the assembly the type is in has to be present, even for SqlClient (which is a common problem, as users think they don't need to reference it like on .net full). 

Your code uses invariant names to obtain the factory from DbProviderFactories on .net full. These invariant names, like System.Data.SqlCient or Npgsql, are in general the namespace of where the DbProviderFactory derived type is located. I don't know of any provider which doesn't use that system for the invariant name (DbProviderFactories uses that scheme as a fallback as well). But in any case: your code uses name X, the name the ADO.NET provider writer wants to use, like Npgsql or System.Data.SqlClient, to obtain the factory. The user, who wants to register the factory, then needs to use that same name X, so the factory is registered under the name your code will use. that's how it also works on .NET full. You can make the registration burden for users a bit easier to simply only require to specify a type (like we do) and derive X from the namespace of the type (as they all use that).

jbogard:
I'd be happy to - may I recommend MyGet to push your CI-level packages to :)

erikbra:
Lovely, @jbogard. We'll probably push CI-packages to MyGet eventually, but we're not doing too much now before @ferventcoder has set up the proper build system. Right now it's only a couple of powershell scripts, and we're planning on using Cake. I think it might be easier to push CI packages if we use a `develop` branch, people. Because then GitVersion produces more sensible -alpha (or -unstable in GitVersion 3.x) and -beta version numbers. Right now every package which is not a pull request looks like a release because of the "clean" version number it gets from being build from `master`.

erikbra:
@FransBouma Thanks for the explanation. So, this is like a ServiceLocator pattern, with names registrations, am I right? Is the common use case to take the provider name from a Connection String, and then return a factory? I see we can solve it differently in RoundhousE, as we have a subclass per provider. But, I was wondering how people use this now in .NET Desktop. Are there no lists in the registry, etc in full Framework either?

FransBouma:
@erikbra in .net desktop, factories are registered in machine.config, or if needed in the app/web.config file. Most ado.net providers still come with an installer which registers a line in machine.config, which section is read by DbProviderFactories to build its datatable with name-factory instances. 

On .net core there's no machine.config so your code can't locate the DbProviderFactory instance. It therefore needs to be registered somewhere, by the dev using your code. This is where your registration method comes in.

erikbra:
Yes, that's what I thought, @FransBouma. That there was some kind of central registry. But there is no such concept in .net core. And of course, then, they need to register it. I just wonder, is it a very common use case to have a string (that is effectively the full name of the provider), and want a factory? I'm just wondering, as I'm not familiar with the scenarios.

erikbra:
I'll write some tests on this PR, to see that we get the same Factory implementations for each type before and after the conversion.

erikbra:
Added unit tests (before conversion to netstandard2.0), all tests green except Access - couldn't understand how to get the invariantProviderName to work). Converted to new strategy - with "hard-coding" the provider in each of the database-specific database implementations, and then all tests ran green.

@jbogard - could you please test your usecase with netstandard projects (you use the roundhouse.lib nupkg, don't you), d/l from here: https://ci.appveyor.com/project/chucknorris/roundhouse/build/0.8.9-PullRequest.294+34.28/artifacts

@ferventcoder and @BiggerNoise - do we want to merge this one? I think at least we should wait until we get the `EnvironmentName` vs `EnvironmentNames` sorted out (#296). Do we have some strategy for doing testing on all the providers?

jbogard:
I'm getting an exception when doing a related reflection-based startup (I may need to fix that other library), but I get a `ReflectionTypeLoadException` for these assemblies:

 - `System.Configuration.ConfigurationManager, Version=4.0.0.0`
 - `System.ServiceModel, Version=3.0.0.0`

It's in a unit test project that targets `netcoreapp2.0` and the exception comes from me looping over the AppDomain's loaded assemblies. Command line is a little different (it runs a different runner for xunit):

```
      System.TypeInitializationException : The type initializer for 'ContosoUniversity.IntegrationTests.SliceFixture' threw an exception.
      ---- System.IO.FileNotFoundException : Could not load file or assembly 'System.Configuration.ConfigurationManager, Version=4.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. The system cannot find the file specified.
      Stack Trace:
        C:\dev\ContosoUniversityDotNetCore\ContosoUniversity.IntegrationTests\SliceFixture.cs(45,0): at ContosoUniversity.IntegrationTests.SliceFixture.ResetCheckpoint()
        C:\dev\ContosoUniversityDotNetCore\ContosoUniversity.IntegrationTests\IntegrationTestBase.cs(20,0): at ContosoUniversity.IntegrationTests.IntegrationTestBase.<InitializeAsync>d__2.MoveNext()
        --- End of stack trace from previous location where exception was thrown ---
           at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
           at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
        --- End of stack trace from previous location where exception was thrown ---
           at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
           at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
        --- End of stack trace from previous location where exception was thrown ---
           at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
           at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
        ----- Inner Stack Trace -----
           at System.Reflection.RuntimeAssembly.GetExportedTypes(RuntimeAssembly assembly, ObjectHandleOnStack retTypes)
           at System.Reflection.RuntimeAssembly.GetExportedTypes()
           at System.Linq.Enumerable.SelectManySingleSelectorIterator`2.MoveNext()
           at MediatR.ServiceCollectionExtensions.AddMediatRClasses(IServiceCollection services, IEnumerable`1 assembliesToScan)
        C:\dev\ContosoUniversityDotNetCore\ContosoUniversity\Startup.cs(51,0): at ContosoUniversity.Startup.ConfigureServices(IServiceCollection services)
        C:\dev\ContosoUniversityDotNetCore\ContosoUniversity.IntegrationTests\SliceFixture.cs(38,0): at ContosoUniversity.IntegrationTests.SliceFixture..cctor()
```

erikbra:
Thanks for testing, @jbogard. Is this maybe the cause of your `System.Configuration.ConfigurationManager` problem? Or maybe not. This is when targeting net461: https://stackoverflow.com/questions/46360716/cant-use-system-configuration-configuration-manager-in-a-net-standard2-0-libra

I see we're building against an ancient NHibernate. That may be a reason...

jbogard:
Yeah I think so, those ancient versions because of the shims bring in all of .NET Framework

BiggerNoise:
If we nuke NHibernate (as we have discussed elsewhere), would that simplify this task?

jbogard:
So one thing I would make sure is for your dependencies, make sure they're .NET Standard also, not full .NET.

erikbra:
I updated all the nuget refs to newest version (bumped NHibernate two major versions), let's hope it betters things. Do you have a public repo you're testing on, @jbogard? Or a minimal repro of the error? So that I could test too. 

But this might not be us. I know you saw this one too... https://github.com/dotnet/standard/issues/567

jbogard:
Yep, it's a branch for this PR https://github.com/jbogard/ContosoUniversityDotNetCore/pull/3

I had to pull the package down because I don't see a NuGet feed for the AppVeyor artifacts.

erikbra:
I'll have a look at the code, @jbogard 
btw, there's a nuget feed off AppVeyor  here: https://ci.appveyor.com/nuget/roundhouse

erikbra:
I found the problem, and it's with NHibernate, which doesn't work with .NET Core yet. They are very close, and I think version 5.1 will work on core.

@jbogard Linq is great when everything goes as expected. When something unexpected fails, it's not that easy to find errors. I had to download and change AutoMapper.Extensions.Microsoft.DependencyInjection a bit to find the error. And I found that it was only NHibernate that was a problem.

I changed `ServiceCollectionExtenstions.AddAutoMapperClasses` (lines 70ff) to this to find the assembly that was the problem.

```
  List<TypeInfo> allTypes = new List<TypeInfo>();

            foreach (var a in assembliesToScan)
            {
                try
                {
                    if (a.GetName().Name != nameof(AutoMapper))
                    {
                        allTypes.AddRange(a.DefinedTypes);
                    }

                }
                catch (Exception e)
                {
                    Console.WriteLine("Problems with assembly " + a.FullName + ": "+ e);
                }
            }
```

I cheched out the NHibernate repo, and found out they weren't quite ready for .NET Core yet, but very close: https://github.com/nhibernate/nhibernate-core/pull/633, I think they're good in the next release. I then changed to multi-target net461 and netstandard2.0 in roundhouse.lib, and depended on a CI build of NHibernate for netstandard2.0, and everything seems to work. I don't like depending on something that fragile, but it indicates that roundhouse can actually run on netstandard2.0. 

I got *some* green tests, @jbogard, but some red, but I don't know how much setup you would expect to do to run all tests green. Would you mind testing with the latest CI build of this pull request? I still get some errors, but I hope it's not RoundhousE.lib's fault, but the setup lacking something:

> A type could not be created from the object you passed. "roundhouse.databases.sqlserver.SqlServerDatabase, roundhouse.databases.sqlserver" resolves to null.

Could you please check out if the RoundhousE part of things work now, with this build? https://ci.appveyor.com/project/chucknorris/roundhouse/build/0.8.9-PullRequest.294+37.33/artifacts

jbogard:
So is this going to wait for NH 5.1, or going to remove the NH dependency altogether?

erikbra:
Good question, @jbogard. Any thoughts on this, @ferventcoder or @BiggerNoise ? Could we manage without NHibernate? What are we using it for? Is it for creating the RoundhousE.* tables?

BiggerNoise:
Yes, We use NHibernate for managing the RoundhousE.* tables. I would expect that a move to Entity Framework would be pretty low effort and a move to just not use this sort of library wouldn't be all that much work either. I know EF has some issues with core, but I thought they were fairly manageable.

We should solve #16 while we deal with this.

jbogard:
Or just ADO.NET :)

EF doesn’t support .NET Core. And EF Core isn’t .NET Standard.

ferventcoder:
Honestly the tables change so often ... ;)

ADO.NET would prolly be sufficient, I was just not a fan of wiring up all of that code to compare the model to the table and then generating the code to make the changes (per db type) when there was already a tool that was sufficiently doing this (barring #16).

BiggerNoise:
We could probably manage it ourselves. I have a couple of thoughts in that area. Before I go too far down that rabbit hole, I wanted to revisit @jbogard and the comment that EF doesn't support .NET core.

I was referring to EF.Core which certainly [appears to be supported](https://docs.microsoft.com/en-us/ef/core/platforms/). I don't know if it has the migration features to make it worth using to handle migrating RH's tables. If not, then there's certainly no point; everything else RH does could be easily handled by plain ole ADO.NET.

Of course, Oracle has to [throw a wrench in the works](https://docs.microsoft.com/en-us/ef/core/providers/oracle/) and not be supported under EF.Core. It looks like they may not even be supporting .NET Core at all. https://blogs.msdn.microsoft.com/dotnet/2016/11/09/net-core-data-access/

Can we vote them off the island?

ferventcoder:
@BiggerNoise you do understand doing that with ADO.NET could easily double the current code set? Just wanted to be sure I threw that out there. I would also hope for some tests catching scenarios that are not easily reproducible, like #16.

ferventcoder:
Oracle (I read the comments :/) is maybe planning support by the end of 2017 - http://www.oracle.com/technetwork/topics/dotnet/tech-info/odpnet-dotnet-core-sod-3628981.pdf

BiggerNoise:
I wouldn't expect that merely inserting the records to track the migrations that we have run would double things up, However, trying to compare to an ideal model and migrate our way there (across multiple database engines) kinda gives me the willies.

If we had to just do ADO.NET, I would recommend that we do something like (11:30 PM rough draft caveat):
1. Embed our migration scripts as resources in the DLL.
2. Maintain a very, very simple migrations run table in the roundhouse schema (more along the lines of what Rails creates).
3. Run our migrations against this simple table when RH initializes.

That gets us out of the business of trying to write a cross-platform schema migrator and returns us to a fairly straightforward logistical problem. It's probably complicated by cross-platform differences in the simple migration scripts, but that's probably still considerable simpler to deal with than trying to detect and adjust the schema.

My list of preferences would probably look something like this:
1. Stick with FNH and they magically get the .NET Core version ready to roll
2. Work with EF CORE which has all the capabilities we need for our simple migrations
3. Something like what I have described here

Of course, I'm assuming that kicking Oracle off the island isn't really an option which seems to put this entire effort on hold.

ferventcoder:
There is an option for a temporary kick, which means point folks needing oracle to stick with an older version until support is added (and pressure Oracle to get it done).

ferventcoder:
Migration scripts embedded is absolutely what we would need, for each db type. The trick is ensuring the schema and table names are changeable at runtime.  That should be enough per type to get them  run. Tracking though? It's like we need a scripts run rable for RH itself.

ferventcoder:
Also 11:30PM reading... you stated pretty much the same thing. 😂

BiggerNoise:
Yes. It would still be a pain though.

BiggerNoise:
This why I don't commit code after midnight; it's like feeding the gremlins, nothing good can come from it.

erikbra:
I had the same thougts. It feels a bit counter-intuitive that a DB change management tool needs another DB change management tool to manage its own tables. I really like the notion of dogfooding, an using RH to manage RH (this could of course lead to infinite recursion - how would we manage changes to the meta-RoundhousE tables?) - but maybe with a much simpler format, yes.

Ref @jbogard s comment aboudn EF Core. EF.Core supports netcoreapp2.0, not netstandard2.0.
See any of these for details of the difference:

https://docs.microsoft.com/en-us/dotnet/standard/net-standard
https://msdn.microsoft.com/en-us/magazine/mt842506.aspx


So, we could, if we find a way to be netstandard-compatible in the libs, target both netcoreapp2.0 and net461 on rh.exe, even if the libs are netstandard2.0.

kieranbenton:
Have you guys got any more thoughts on this or is there anything I can do to help out?

erikbra:
Good question, @kieranbenton . I haven't really had time to look more into it. I think writing out the dependency on NHibernate would be the most reasonable long-term solution. I have tried to follow NHibernate's road to .net core 2.0 /netstandard 2.0, and it is also a road of many unfinished pull requests and a lot of discussion.

So, if we could manage to write out the NHibernate dependency, that would be ideal. Whether this should be the PR for this or not, I'm not sure. But this definitely depends on it for now (until NHibernate manages to get to core/standard).

@BiggerNoise do you have any more thoughts on you micro-rh for managing the three RoundhousE tables? I really liked the idea. Creating the tables could be done in ANSI SQL, probably, so that would work. But changing is probably much harder, as the informational schemas/tables/views of each DB systems are quite different. 

How common is change in the RH tables, anyway? Are we trying to solve a non-issue? Just asking. It might be a real issue.

BiggerNoise:
My thoughts on this are twofold:
1. It's a totally worthwhile effort that would eliminate a significant dependency
2. It's going to be an epic pain in the butt

The tables change very infrequently, but they do change.

It looks like EF Core is [now supporting .NET Standard 2.0](https://docs.microsoft.com/en-us/ef/core/what-is-new/), so that avenue looks like it would be open to us.

I'll try and spike a PoC using EF.Core to manage this migration in the next week or so.

BiggerNoise:
I have created a small sample [here](https://github.com/BiggerNoise/ef-core-poc).

It took a lot longer to get to than expected, so it's just doing one table in SQL server. I'll refactor it to also support postgress and that should give us an idea what this might look like if we were going to head in that direction.

So far, it's promising, but I am still wondering if we should just put our maintenance scripts in resources and just run them. It may ultimately be less painful.

BiggerNoise:
I have updated the sample to better reflect how we might use this. I think it's actually very promising.
There's four projects in the folder:

A base that defines a context and a scripts run model
A sql context with sql server migrations
A pg context with Postgresql migrations
A command line tool

To keep things simple, I just checked in the pw that I use for my local (no remote access) postgres instance. You'll probably need to change it for your own use. Please make sure that you don't accidentally commit creds for an important database.

in the cmdline folder, you run things with:
```
dotnet run -- mssql
```
Or

```
dotnet run -- pg
```

To add a new migration, you change the model and then go to each of the context projects and add a migration

```
$ cd ../ef-core-poc-sql/

$ dotnet ef migrations add AddScriptsRun_run_action --startup-project ../cmdLine/
Done. To undo this action, use 'ef migrations remove'

$ cd ../ef-core-poc-pg/

$ dotnet ef migrations add AddScriptsRun_run_action --startup-project ../cmdLine/
Done. To undo this action, use 'ef migrations remove'
```

The most recent commit is the addition of the run_action field to scripts run and the associated migration.

If this seems like it will work for others, I can add a mysql migrator as well.

BiggerNoise:
We actually use Fluent Nhibernate so depend upon that library. 

It looks like there's a lot of interest in supporting .NET Standard over that way, so we probably will not need to wait too much longer.

hpbieker:
Will this PR make roundhouse available as a dotnet core app, or just make the libraries available as netstandard?

I would like to run roundhouse in a docker compose file on Linux.

BiggerNoise:
Our intention is to make roundhouse available as a dotnet core app. I don't think this P/R will get us all the way there, though. It is, however, a necessary step.

cfletcher:
Looks as though FluentNHibernate is now released: https://github.com/FluentNHibernate/fluent-nhibernate/releases

BiggerNoise:
Thank you. I'm out at NG Conf this week, so there's probably not going to be much motion on this this week.

This has been hanging around for a while, so I'll try to get into it next week.

erikbra:
Just wanted to say I'm sorry I haven't been able to involve myself here for ages. Really demanding times at my day-job :)

Lovely to see you are taking ownership of this, @BiggerNoise . I'll see what I can do going forward, but the next month or two are probably going to be killing me here as well...

BiggerNoise:
Hey all-

I know there's a lot of interest in this one, but it's going to be a while before I can spend any time on it.

My mom is in the Hospital, and I'll be shuffling back and forth a bit between Texas and Virginia for a while. I don't expect to have time or attention to give this the attention that it will need.

-Andy

BTW - She is receiving excellent care and her prognosis is good.

erikbra:
The latest commits (rebased, sorry, should've merged - looks like everything is new commits)
 are not ready to merge, just a WIP.  Had a little go at it, upgraded NHibernate, etc. Doesn't look like ILMerge is very fond of netstandard libraries, doesn't work properly yet.

erikbra:
This is fun. We have a few problems with the dependencies being ancient, so we have some upgrading to do to get there. Now only SQL Server works on .net core (the rest are conditionally included when on net461). I haven't actually run on core against a database yet, but I must say this is pretty satisfying in itself ;)

On a mac...
image

erikbra:
The nuget packaging isn't working either. Guess we'll have to switch to `dotnet pack`, but that should be OK. Getting there bit by bit...

BiggerNoise:
Thanks @erikbra for poking this one.

I believe that I saw a reference to another tool that would replace ILMerge and it handles the core run time. I can't seem to find it now, but I'll attempt to take a further look tomorrow.

erikbra:
Lovely. That said, I now successfully run `dotnet bin\netcoreapp2.0\rh.dll` on a 300+ RH scripts solution, accumulated over a couple of years. And it seems to work nicely. Ran on Windows, though, with SQL server. Haven't tested anything else. When I find the time, I will test on SQL server on Linux on macOS (in a container). I don't have any other database projects to test MySql, PostgreSQL, Oracle, etc. It would be nice to test on some real-world project. Do you have any tips?

Btw, hope your mother is doing OK given the circumstances.

erikbra:
There are actually a couple of alternatives for merging/linking on netcore. I don't have experience with any of them, but they're worth checking out.

There's at least:

* The CoreRT project (https://github.com/dotnet/corert)
* ILLinker (https://github.com/dotnet/announcements/issues/30)
* Open-sourced ILMerge itself: https://github.com/Microsoft/ILMerge

We should evaluate them all, and choose the best one for our needs. A single binary for all OS-es (same or different, but at least _one_ binary, not a bunch of dlls) would be nice.

cfletcher:
If you're working on cross platform compatibility Scott Hanselmann referenced a Roslyn analyser that helps with this kind of thing https://www.hanselman.com/blog/WritingSmarterCrossplatformNETCoreAppsWithTheAPIAnalyzerAndWindowsCompatibilityPack.aspx

erikbra:
Thanks, @cfletcher . I'll have a look at it. I suspect some of the libraries used are so old the tool won't work, and I don't think it fixed e.g. path separator slash directions. But, it might definitely give us some advice.

erikbra:
@BiggerNoise I think this one's done now. But I am in lack of testing environments, and don't have the knowledge of MySql, PostgreSQL, Oracle, etc, to test. 

I have now made two "command line" distributions of RoundhousE, one as was, and one for .net core. I have tried documenting the usage in the `README.markdown` file. I am very open to discussing what ways we should distribute RoundhousE for non-windows platforms

I have also tried to make the roundhouse.lib nuget package multi-targeted, i.e. for both netstandard2.0 and net461. I hope I didn't break anything.

I wasn't able to make the Oracle and Access providers to compile on netstandard, so those are excluded when building the netcoreapp2.0 one. But the other ones compiled, at least. And I have tested the SqlServer one rather properly on macOS, running against SQL Server in a docker container (imagine that).

There are two major changes:
1. Conversion of most code to .netstandard2.0
1. Swapped out Microsoft.EnterpriseLibrary.TransientFaultHandling with Polly, and some custom code. I hope someone would like to test this a bit. I have done my best, but you never know ;)

Testing help is much appreciated, as people have so many different usages and environments they use RoundhousE in.

@jbogard , I know you have been eager to try this, so if you could please give it a shot and see if it works as expected, I would really appreciate it.

BiggerNoise:
@erikbra Thank you so much for this. I should be able to beat on this a bit later in the week.

I am thinking that this is probably a good point to put a stake in the ground and call this 1.0. Can we start cranking out 1.0 release candidates without merging to the master branch?

erikbra:
I guess the best way to go branch, wise, would to create a branch off master in chucknorris/roundhouse, e.g. "release-1.0" or something, and make this PR go in there instead. Whether I can change this PR from erikbra/roundhouse (branch netcor) to go against the new branch in chucknorris/roundhouse or I have to create a new one, I don't know. But we'll do what we can.

I guess we can release some RCs without merging to master. Maybe we need to adjust the appveyor.yml a bit, maybe not. Maybe we should put them on Myget? Or is Nuget fine for release RCs as well?

I guess we should make a .deb for distribution to *nixes eventually, but that should be a separate PR, I guess. If we make a release branch, I can see if I can look into it. Read a bit about .debs in the weekend, I have no prior experience, but it doesn't look impossible. I think we should set up a build step on a linux machine, though, so it should probably be a separate project in AppVeyor (you can have multiple .yml files).

BiggerNoise:
I can manage getting the new branch and getting this P/R merged into that one. My main concern is opening up the testing audience to people that don't feel like building RH from source.

I agree that proper distribution to Linux and MacOS should be separate P/Rs. I'd be willing to look at those once we have a solid 1.0 out the door.

erikbra:
Yes, of course you should be able to test without compiling. That's why I created a .tar.gz file in the artifacts. This can be `tar -zxf`'d to anywhere on a *nix box with dotnet core installed, and run. I wrote about it in README.markdown, but it doesn't seem like you can view the PR's on the main page. You can read it here, anyway (in my fork):

https://github.com/erikbra/roundhouse/tree/netcore#dotnet-core-linux-macos-etc

Is that a "good enough" experience for beta testers, do you think? If we upload the .tar.gz to the "Releases" page here on GitHub?

BiggerNoise:
I hope that I am not making an absolute hash of this, but I have:

- Added a release/1.0.0 branch
- Merged this P/R into that branch
- Tagged with 1.0.0.RC1

Huge thanks to @erikbra for getting this together.

erikbra:
Great work, @BiggerNoise.

Always exciting to do something remotely unfamiliar with git. :) I always get burned now and then. 

I'm not sure if our GitVersion configuration understood the release branch name, it still marks it at 0.9.2-pr-something. I think there's somewhere you can tell it the format of a release branch (i.e. `release-1.0.0` vs. `release/1.0.0`, etc)

I don't have much experience working with PRs a lot on the "receiving side". But, say I was to make a bugfix or something on the release branch? How would I do that? Can I just push to this PR, or is that link "disconnected" now?

BiggerNoise:
I would open a new P/R from the 1.0.0 branch.

erikbra:
I was looking at the wrong build in AppVeyor, it does understand the branch names. It just didn't respect the rc tag, it looks like. It's naming it 'beta', but 1.0.0 is picked up:

https://ci.appveyor.com/project/chucknorris/roundhouse/build/1.0.0-beta.1+22.build.95

Maybe the tag should have been 1.0.0-rc.1, with a dash instead of a dot btw the version and the suffix? The docs are  a little unclear, but it could look like that should work: https://gitversion.readthedocs.io/en/latest/examples/

BiggerNoise:
I just pushed with 1.0.0-rc1, let's see how that works out

erikbra:
The version number got picked up by GitVersion. The package versions are now OK. I made an attempt at updating the instructions on how to get the preview packages. I couldn't get the package to show up in the public browsing part on MyGet, it might be something in the security settings, but I couldn't find it. However, the Nuget package feed is publicly available.

cfletcher:
@erikbra I'm looking for package on the nuget feed but I can't see it, am I missing something? 
https://www.nuget.org/packages/roundhouse/
The most recent package is 0.9.1 from January

erikbra:
@cfletcher sorry, we haven't pushed it to NuGet yet. We do have some issues still, but I think that's mostly with the `rh.exe` command-line utility. The RCs are available from 
https://www.myget.org/F/roundhouse/api/v3/index.json

Unfortunately, neither of @BiggerNoise or I have had enough time to work on this the past months, so we haven't got the final 1.0.0 out yet. I think this one is a blocker: #327

iainnicol:
With that blocker now fixed, would you consider uploading an RC to nuget.org?  Either way, thanks for your work on this.

erikbra:
I haven't got stuff out on nuget yet, and I'm having a bit of trouble with the packages built on AppVeyor, so the MyGet auto-upload isn't any good for the dotnet tool. But the libs, etc, should be good. 

Just add this to your NuGet feed if you are impatient, or wait a couple of days (hopefully), until we get it out on NuGet.

https://www.myget.org/F/roundhouse/api/v3/index.json

iainnicol:
I'm not sure it's obvious, but the v1.0.0-rc packages _are_ uploading to MyGet.  See https://www.myget.org/feed/roundhouse/package/nuget/roundhouse/1.0.0-rc.2 for an overview. The download link on that page works.

Surprisingly, the v1.0.0-rc packages don't show appear in the index.json. I would guess MyGet privacy settings, or else a MyGet bug, not Appveyor's fault.

I saw in another thread you uploaded (manually?) to NuGet. Great! Thanks.

erikbra:
It's maybe not obvious. But they are on MyGet, yes. But the dotnet tool on MyGet is not working. Don't know why. But, now they're also all live on NuGet. E.g: 
https://www.nuget.org/packages/dotnet-roundhouse/ (the dotnet global tool)

https://www.nuget.org/packages?q=roundhouse (all of them)

erikbra:
Don't know if you're still waiting around for this, @jbogard, but the RCs of RoundhousE for .net core are up on NuGet now.

jbogard:
Nice!

PR #295, opened 2017-10-29
Removed obsolete properties, RecoveryModeSimple and EnvironmentName

erikbra:
(Completes #293 in a better way).

PR #296, opened 2017-10-30
Revert too eager deletion of the EnvironmentName config setting

erikbra:
This reverts a bit too eager deletion of the `EnvironmentName` configuration property, from pull request #295.

### DIFF
--- a/product/roundhouse/infrastructure.app/ApplicationConfiguraton.cs
+++ b/product/roundhouse/infrastructure.app/ApplicationConfiguraton.cs
@@ -124,7 +124,10 @@ namespace roundhouse.infrastructure.app
             }
             if (string.IsNullOrEmpty(configuration_property_holder.EnvironmentNames))
             {
-                configuration_property_holder.EnvironmentNames = ApplicationParameters.default_environment_name;
+                if (!string.IsNullOrEmpty(configuration_property_holder.EnvironmentName))
+                    configuration_property_holder.EnvironmentNames = configuration_property_holder.EnvironmentName;
+                else
+                    configuration_property_holder.EnvironmentNames = ApplicationParameters.default_environment_name;
             }
             if (string.IsNullOrEmpty(configuration_property_holder.OutputPath))
             {
--- a/product/roundhouse/infrastructure.app/ConfigurationPropertyHolder.cs
+++ b/product/roundhouse/infrastructure.app/ConfigurationPropertyHolder.cs
@@ -40,6 +40,8 @@ namespace roundhouse.infrastructure.app
         string VersionTableName { get; set; }
         string ScriptsRunTableName { get; set; }
         string ScriptsRunErrorsTableName { get; set; }
+        [Obsolete("Use EnvironmentNames")]
+        string EnvironmentName { get; set; }
         string EnvironmentNames { get; set; }
         bool Restore { get; set; }
         string RestoreFromPath { get; set; }


BiggerNoise:
Question for @ferventCoder: Is the configuration property part of the API surface area or is it something that we just use internally.

If it's just internal, then there is no need for this P/R to revert the changes.

erikbra:
Looks like we can actually skip this "repair" pull request, yes, @BiggerNoise. In `Program.cs`, the property actually assigned to, is `ConfigurationPropertyHolder.EnvironmentNames` (plural).

image

There are also tests on this (not testing the command-line switches, but the configuration properties themselves): 

        [Concern(typeof(DefaultDatabaseMigrator))]
        public class when_determining_if_we_are_in_the_right_environment_with_multiple_environments : concern_for_database_migrator_with_multiple_environments

(from `DefaultDatabaseMigratorSpecs.cs`)

But I would be more comfortable is @ferventcoder could confirm it too :)

product/roundhouse/infrastructure.app/ApplicationConfiguraton.cs
```diff
@@ -124,7 +124,10 @@ public static void set_defaults_if_properties_are_not_set(ConfigurationPropertyH
             }
             if (string.IsNullOrEmpty(configuration_property_holder.EnvironmentNames))
             {
-                configuration_property_holder.EnvironmentNames = ApplicationParameters.default_environment_name;
+                if (!string.IsNullOrEmpty(configuration_property_holder.EnvironmentName))
+                    configuration_property_holder.EnvironmentNames = configuration_property_holder.EnvironmentName;
+                else
+                    configuration_property_holder.EnvironmentNames = ApplicationParameters.default_environment_name;
```
ferventcoder: No matter what, this section needs to use the new name - think this was an oversight in the original PR (that moved to env names)

product/roundhouse/infrastructure.app/ConfigurationPropertyHolder.cs
```diff
@@ -40,6 +40,8 @@ public interface ConfigurationPropertyHolder
         string VersionTableName { get; set; }
         string ScriptsRunTableName { get; set; }
         string ScriptsRunErrorsTableName { get; set; }
+        [Obsolete("Use EnvironmentNames")]
+        string EnvironmentName { get; set; }
```
ferventcoder: This may be able to be set from the RoundHousE api, so best to check if you can property config in first

ferventcoder:
Program.cs is when it sets command line only values. With RH you need to consider other client access, lile the rh api, or the MSBuild task.

product/roundhouse/infrastructure.app/ConfigurationPropertyHolder.cs
erikbra: I didn't quite understand what you mean. I have effectively just reverted to "old" (as in original EnvironmentName -> EnvironmentNames PR) behaviour. Should we make "EnvironmentName" a Facade for EnvironmentNames instead, maybe? (i.e. write-through on set, read from EnvironmentNames on read?)

ferventcoder: I think EnvironmentName is effectively a facade now

ferventcoder: it could probably use a little bit more to make that change effective.

erikbra: Yes, maybe. So we might as well just make it a proper facade in `ConfigurationPropertyHolder`, and move the "if empty then the other one, else this" logic there, or remove it altogether (as in the original trigger-happy PR). What do you prefer? 

I prefer making `EnvironmentNames` a list (of some sorts) of string, and make conversion (split on comma or whatever) on reading command-line. It makes more sense when working with RoundhousE as a library. Unbound strings are evil. You can store anything from a single character to the complete works of Shakespeare in them, and there's no way of telling from the signature if there is a "hidden format" you should use (e.g. comma-separated strings).

ferventcoder: I prefer one environment name, but it should be converted (and I thought the original PR adding this handled the splitting?) to a list.

ferventcoder: There may be a reason the others are all strings here though. Check to see if there is a list type in property configuration holder - if not, there may be a reason for it.

erikbra: I will have a closer look.

BiggerNoise: Agree that the property should be a string collection and that it should be split when the command line is read. Parsing a comma separated string provided as an API argument is just silly.

We can have the obsoleted singular property that is just a facade for the collection; agree with that approach.

All of this code is present, just a matter of some reorganization. I'm busy tonight, but I can definitely put this together Wednesday night. If you want to do this one, feel free, but I definitely don't mind jumping in on some code (build systems bore me to tears...yes, they're really important, but I'd rather watch a wall of paint dry.)

erikbra: Lovely if you can have a look, @BiggerNoise. I'd feel more comfortable on this one if we have more than two eyes on it, as we've had some discussions around it. Can you technically commit to/continue on the same pull request, how does that work?

erikbra: Perfect. I Agree, if we have many people working on the same thing, a branch is better. But not necessary here.

BiggerNoise:
Closing in favor of #297 which includes this work as well as the changes to the implementation discussed here.

PR #297, opened 2017-11-02
Erikbra fix too eager delete

BiggerNoise:
Includes and supersedes #296 

EnvironmentName is restored (and marked obsolete), but is now implemented in terms of EnvironmentNames and the splitting is done immediately.

### DIFF
--- a/product/roundhouse.console/Program.cs
+++ b/product/roundhouse.console/Program.cs
@@ -28,7 +28,7 @@ namespace roundhouse.console
 
     public class Program
     {
-        private static readonly char[] OptionsSplit = new[] { ',' };
+        private static readonly char[] OptionsSplit = new[] { ',',';' };
 
         private static readonly ILog the_logger = LogManager.GetLogger(typeof(Program));
 
@@ -259,7 +259,7 @@ namespace roundhouse.console
                 //environment(s)
                 .Add("env=|environment=|environmentname=|envs=|environments=|environmentnames=",
                      string.Format(
-                         "EnvironmentName(s) - This allows RH to be environment aware and only run scripts that are in a particular environment based on the naming of the script. LOCAL.something.ENV.sql would only be run in the LOCAL environment. Multiple environments may be specified as a comma-separated list. Defaults to \"{0}\".",
+                         "EnvironmentName(s) - This allows RH to be environment aware and only run scripts that are in a particular environment based on the naming of the script. LOCAL.something.ENV.sql would only be run in the LOCAL environment. Multiple environments may be specified as a comma or semicolon separated list. Defaults to \"{0}\".",
                          ApplicationParameters.default_environment_name),
                      environmentOption =>
                      {
--- a/product/roundhouse.console/Program.cs
+++ b/product/roundhouse.console/Program.cs
@@ -28,6 +28,8 @@ namespace roundhouse.console
 
     public class Program
     {
+        private static readonly char[] OptionsSplit = new[] { ',' };
+
         private static readonly ILog the_logger = LogManager.GetLogger(typeof(Program));
 
         private static void Main(string[] args)
@@ -259,7 +261,13 @@ namespace roundhouse.console
                      string.Format(
                          "EnvironmentName(s) - This allows RH to be environment aware and only run scripts that are in a particular environment based on the naming of the script. LOCAL.something.ENV.sql would only be run in the LOCAL environment. Multiple environments may be specified as a comma-separated list. Defaults to \"{0}\".",
                          ApplicationParameters.default_environment_name),
-                     option => configuration.EnvironmentNames = option)
+                     environmentOption =>
+                     {
+                         foreach (var environment in environmentOption.Split(OptionsSplit, StringSplitOptions.RemoveEmptyEntries))
+                         {
+                             configuration.EnvironmentNames.Add(environment);
+                         }
+                     })
                 //restore
                 .Add("restore",
                      "Restore - This instructs RH to do a restore (with the restorefrompath parameter) of a database before running migration scripts. Defaults to false.",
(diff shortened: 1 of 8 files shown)


product/roundhouse.tasks/Roundhouse.cs
```diff
@@ -110,9 +112,15 @@ bool ITask.Execute()
         public string ScriptsRunErrorsTableName { get; set; }
 
         [Obsolete("Use EnvironmentNames")]
-        public string EnvironmentName { get; set; }
-
-        public string EnvironmentNames { get; set; }
+        public string EnvironmentName {
+            get { return EnvironmentNames.SingleOrDefault(); }
```
erikbra: This will crash if you have more than one environment in the list. That's probably OK? We don't have a better way of solving it, do we?

erikbra:
This looks  good to me, @BiggerNoise. But should we bring in @ferventcoder  to merge, since both you and I have worked on it? Or should one of us just merge it?

product/roundhouse.tasks/Roundhouse.cs
BiggerNoise: I would say the exception is the desired behavior. Using a mixture of EnvironmentName and EnvironmentNames is incorrect. I like to fail loudly when being mis-used.

BiggerNoise:
I'd like @ferventcoder  to do the merge if he could.

product/roundhouse.tasks/Roundhouse.cs
erikbra: I agree. Leave it as is :)

product/roundhouse.console/Program.cs
```diff
@@ -28,6 +28,8 @@ namespace roundhouse.console
 
     public class Program
     {
+        private static readonly char[] OptionsSplit = new[] { ',' };
```
ferventcoder: Split on semi-colon as well.

product/roundhouse/infrastructure.app/ConfigurationPropertyHolder.cs
```diff
@@ -40,7 +40,9 @@ public interface ConfigurationPropertyHolder
         string VersionTableName { get; set; }
         string ScriptsRunTableName { get; set; }
         string ScriptsRunErrorsTableName { get; set; }
-        string EnvironmentNames { get; set; }
+        [Obsolete("Use EnvironmentNames")]
+        string EnvironmentName { get; set; }
+        IList<string> EnvironmentNames { get; }
```
ferventcoder: As this is part of the API now, what may this possibly break?

ferventcoder: The change from a string to a list, that is.

ferventcoder: Also, there may be an issue with .NET 4.x where you need to also have the set here.

BiggerNoise:
EnvironmentNames is a new property from the 0.8.8 release so the breakage should be fairly limited.

Should only affect library users that have used that new feature. Task and command line should be unaffected.

ferventcoder:
@BiggerNoise that's what I was thinking as well.

product/roundhouse/infrastructure.app/ConfigurationPropertyHolder.cs
BiggerNoise: I don't think there's an issue with only having the get in the interface (implementer has a private set). That should be a C# thing, not a .NET thing, and we're back on C# 5 and it works like a champ. I seem to recall doing this while the earth was still cooling, so I don't think it's a new feature.

Can you elaborate on your concern?

BiggerNoise:
Cool, I'll get the extra split into the array and push it up. Should be tonight.

product/roundhouse/infrastructure.app/ConfigurationPropertyHolder.cs
ferventcoder: Sure can - https://github.com/chocolatey/choco/blob/eaedc2b50b8573c8fd269e56921e92e26db93caf/src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs#L229.

http://stackoverflow.com/a/23809226/18475

If we want to reflect over the configuration information and output that for the logs, for the users, and we support .NET 4.0+, on .NET 4.0 it will cause errors.

BiggerNoise:
OK. @ferventcoder, merge away!

product/roundhouse/infrastructure.app/ConfigurationPropertyHolder.cs
BiggerNoise: I notice that configuration class is marked [serializable] and that serialization was referenced in the linked S.O. question. Are you sure that wasn't where the issue cropped up? I have definitely had  issues with privileged sets when dealing with serialization.

I'd rather leave it private for now and open it up if we find there's a use case that we can't accommodate with the privileged set.  Adding access to a set isn't going to break anyone.

BiggerNoise: OK, @ferventcoder , where are we going with this? I'd like to leave the code as is, but will tolerate making the set public. More than anything, I want to get this merged so we can move on.

ferventcoder: Merge away. If you are ready, I would say consider me more an advisor at this point, I'm going to be heads down on an important project (it does involve some RH), so you will see me pop in every once in awhile. 

This project will also be working on Cake as well, so you will see changes that will come back to UC as well.

BiggerNoise: Cool. Didn't want to stomp on an active issue. 

@erikbra if you would do the honors, that would be great.

BiggerNoise:
Thanks

Issue #171, opened 2015-01-29
Switch Oracle to Managed Client


bladeoflight16:
Native binary clients are terrible and problematic. Switch to the full .NET version for Oracle: https://www.nuget.org/packages/odp.net.managed/

PR #85, opened 2012-09-11
Converted all projects to AnyCpu

LodewijkSioen:
This will update all the csproj files to build for 'AnyCpu'. Everything builds and still works on Oracle and MsSql.

PR #298, opened 2017-11-05
{WIP} Changed OracleConnection to use Managed provider

erikbra:
I was tired of the warnings about System.Data.OracleClient being outdated, so I tried to convert to using https://www.nuget.org/packages/Oracle.ManagedDataAccess/ to connect to Oracle databases. In the passage, I created an integration test for Oracle too (made a copy of the MySql one), but I'm having a really, really hard time getting it to work.

Mainly I get a ORA-01017, which is basically "wrong password". However, it is not..
Other people are having the problem too: 
https://stackoverflow.com/questions/14476875/ora-01017-invalid-username-password-when-connecting-to-11g-database-from-9i-clie
https://dba.stackexchange.com/questions/145055/ora-01017-invalid-username-password-logon-denied-from-sql-developer-but-sqlplu

A couple of questions: 
1) Are there any Oracle heads out there who could help me test? I haven't done Oracle in 10 years, and that was from Java
1) Are anyone able to get the new integration test i wrote running with the *current* oracle driver? (it has client dependencies, I didn't want to go down that rabbit hole)

### DIFF
--- a/product/roundhouse.databases.oracle/OracleDatabase.cs
+++ b/product/roundhouse.databases.oracle/OracleDatabase.cs
@@ -1,4 +1,6 @@
-using System.Data.OracleClient;
+using System.Data.Common;
+using Oracle.ManagedDataAccess.Client;
+
 using roundhouse.infrastructure.logging;
 
 namespace roundhouse.databases.oracle
@@ -84,7 +86,12 @@ namespace roundhouse.databases.oracle
 
         public override void set_provider()
         {
-            provider = "System.Data.OracleClient";
+            provider = "Oracle.ManagedDataAccess.Client";
+        }
+
+        protected override DbProviderFactory get_db_provider_factory()
+        {
+            return OracleClientFactory.Instance;
         }
 
         protected override void connection_specific_setup(IDbConnection connection)
--- a/product/roundhouse.databases.oracle/RoundhousEOracleDriver.cs
+++ b/product/roundhouse.databases.oracle/RoundhousEOracleDriver.cs
@@ -1,7 +1,6 @@
-﻿using System.Data.OracleClient;
+﻿using Oracle.ManagedDataAccess.Client;
 using NHibernate.Driver;
 using NHibernate.SqlTypes;
-using roundhouse.infrastructure.logging;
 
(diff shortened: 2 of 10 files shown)


BiggerNoise:
I can't do much more here than say thanks. Hopefully, someone that wants Oracle support is willing to step up and help provide it.

erikbra:
Yes, we'll just have to see. I don't really know anyone using Oracle on .NET these days.

erikbra:
(just if anyone's watching #171 ), this PR tries to address that issue.

@bladeoflight16 - I saw you commented on #85 that we shouldn't switch to AnyCPU without switching to the fully managed Oracle driver. But we have a bit problems testing it. Would you be able to help us get this PR working? I have very rusty Oracle knowledge.

bladeoflight16:
@erikbra Sorry for the delay.

Does the user have permissions to connect? You can make sure they do by running `GRANT CREATE SESSION TO [user]`. (There's also a role that contains the permission, but I personally prefer granting to permission directly, since this makes querying the database's metadata to see if they have it easier.) If the test also creates any objects, you'll need to grant permissions for that as well, and you'll likely need to grant the user some space on the tablespace.

I really wish using Oracle wasn't this complicated.

bladeoflight16:
@erikbra 
Also, if you still want to try to get the existing integration tests to run, I have some advice about the native Oracle client.

You can download the [Instant Client](http://www.oracle.com/technetwork/database/features/instant-client/index-097480.html) (the Basic Light one is all you need) and extract it somewhere. In your command prompt where you're running your tests, you can `SET PATH=%PATH%;C:\path\where\you\extracted\instantclient_12_2` to let RoundhousE to find the binaries. You'll have to do it each time you launch the command line, of course, since that's a process level change to the environment variable.

I personally avoid the global client installs as much as possible, and I've found this works for the majority of applications. (There's more involved if you want to use ASP.NET providers, but RoundhousE doesn't require those.)

erikbra:
I'm abandoning this in favour of grate. I have just now implemented Oracle support there, with the newest Oracle provider. If any of you are interested in having a look, it's here: https://github.com/erikbra/grate/pull/134

PR #300, opened 2017-11-12
Add GitVersion-config

erikbra:
Added GitVersion-config that is supposed to work when we work in a GitHubFlow style (i.e. no separate develop-branch, just PRs off master). Versions from tags, and uses semantic versioning for commits after a release. Uses `pre` as suffix for builds not from a tag.

`master` as of now should be `0.8.9-pre.X`

when we are ready for release, we just tag master with `0.8.9`, and the version number becomes `0.8.9` (no suffix).

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -1,4 +1,4 @@
-#!/usr/bin/env powershell
+#!/usr/bin/env pwsh
 
 $MSBUILD=msbuild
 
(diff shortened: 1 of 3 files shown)


erikbra:
Thanks :)

build.ps1
```diff
@@ -19,7 +19,7 @@ pushd $root
 $gitVersion = (GitVersion | ConvertFrom-Json)
 
 If ($onAppVeyor) {
-    $newVersion="$($gitVersion.FullSemVer).$env:APPVEYOR_BUILD_NUMBER"
+    $newVersion="$($gitVersion.FullSemVer)"
```
ferventcoder: 👍

PR #307, opened 2017-12-14
Fix for IndexOutOfRangeException when token value is empty

alexandergre:
If the token value is empty, roundhouse crashes:
Index was outside the bounds of the array. System.IndexOutOfRangeException: Index was outside the bounds of the array.    at roundhouse.infrastructure.app.tokens.UserTokenParser.<Parse>b__2(String p)    at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)    at roundhouse.infrastructure.app.tokens.UserTokenParser.Parse(String option)    at (...)
Using FirstOrDefault and LastOrDefault instead of accessing array elements directly fixes this.

### DIFF
--- a/product/roundhouse/infrastructure.app/UserTokenParser.cs
+++ b/product/roundhouse/infrastructure.app/UserTokenParser.cs
@@ -23,8 +23,8 @@
 
             if (pairs.Any(p => !p.Contains("="))) throw new FormatException("Wrong format");
 
-            return pairs.ToDictionary(p => p.Split(new string[] { "=" }, StringSplitOptions.RemoveEmptyEntries)[0],
-                p => p.Split(new string[] { "=" }, StringSplitOptions.RemoveEmptyEntries)[1]);
+            return pairs.ToDictionary(p => p.Split(new string[] { "=" }, StringSplitOptions.RemoveEmptyEntries).FirstOrDefault(),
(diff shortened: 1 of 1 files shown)


BiggerNoise:
@alexandergre - Good find and thanks for reporting this.

I added some specs and made the implementation handle another wonky edge case.  That and your commit can be found in commit 966491bc. Thank you again.

BiggerNoise:
@ferventcoder - Did I do the squash right when I merged this P/R? I like the:
- Single Commit
- Detailed Comment

I don't like:
- The original commit SHA is removed from the history so the P/R doesn't automatically close.

Is there a way to get all three?

ferventcoder:
Any changes to a commit became a new SHA. If they were the same it would not mean globally unique.

So there is not a way to have all three unless performed on the original PR.

BiggerNoise:
I figured that was the case (and for those reasons). But, worth asking.

I assume that you would like to have P/Rs merged in this way so that we have those detailed messages about what changed and why we did these things.

ferventcoder:
I prefer merging in the best way possible. If you have really good testing in place, are merging into master, and everything is good to go (requires no changes), it is simple to click "shiny green merge button" directly from here. Otherwise there may be some things to do. Squashing is only necessary to remove commits in the middle that don't really stand on their own, if that makes sense.

Issue #315, opened 2018-04-06
MySql runAfterCreateDatabase not getting executed 


mmalik76:
The MySql runAfterCreateDatabase scripts are not getting executed because the create_database_script() is not returning the "Created" flag like other providers.

mmalik76:
Submitted pull request #316

BiggerNoise:
Thanks for the PR, it's merged and will be part of the next release.

PR #316, opened 2018-04-06
Updated MySql create_database_script to return Created flag

mmalik76:
Fixed issue #315 MySql runAfterCreateDatabase not getting executed

### DIFF
--- a/product/roundhouse.databases.mysql/MySqlDatabase.cs
+++ b/product/roundhouse.databases.mysql/MySqlDatabase.cs
@@ -84,6 +84,7 @@ namespace roundhouse.databases.mysql
                 builder.Database = "information_schema";
                 admin_connection_string = builder.ConnectionString;
             }
+            configuration_property_holder.ServerName = server_name;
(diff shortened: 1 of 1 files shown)


product/roundhouse.databases.mysql/MySqlDatabase.cs
```diff
@@ -109,7 +111,15 @@ private static string build_connection_string(string server_name, string databas
...
         {
             return string.Format(
-                @"CREATE DATABASE IF NOT EXISTS `{0}`;",
+                @"SET @Created = 1;
+
+                SELECT 0 FROM INFORMATION_SCHEMA.SCHEMATA
+                WHERE SCHEMA_NAME='{0}' INTO @Created;
+
+                CREATE DATABASE IF NOT EXISTS `{0}`
+                CHARACTER SET utf8 COLLATE utf8_unicode_ci;
```
BiggerNoise: It looks like this is not the default collation for MySQL. Am I wrong on that?

I'd like to address collation separately and decide if changing the default behavior is a good thing.

PR #319, opened 2018-04-10
Added methods in the Oracle DB section to use the PLSQL methods

RossElder1983:
We were having an issue using roundhouse over numerous schemas on the same oracle instance. 

If we had 2 schemas DB_DEV and DB_TEST, roundhouse would run fine on the DB_DEV instance, create the 3 roundhouse tables and populate them fine. When we came to run then on DB_TEST it would fail stating that the roundhouse tables did not exist. It appeared that it was checking for the existance of the roundhouse tables without taking into account it was under a different schema. 

After a bit of looking about we noticed the PLSQLSpecific class in the roundhouse.databases.oracle section, after re-instating that roundhouse works like a charm.

If this isnt a great idea and there is another way to address the issue we were noticing i'd be very interested to hear some thoughts.

### DIFF
--- a/product/roundhouse.databases.oracle/OracleDatabase.cs
+++ b/product/roundhouse.databases.oracle/OracleDatabase.cs
@@ -102,6 +102,21 @@ namespace roundhouse.databases.oracle
             run_sql(create_sequence_script(scripts_run_errors_table_name), ConnectionType.Default);
         }
 
+        public override void create_or_update_roundhouse_tables()
+        {
+
+            Log.bound_to(this).log_an_info_event_containing("Creating table [{0}].[{1}].", roundhouse_schema_name, version_table_name);
+            run_sql(PLSQLSpecific.create_roundhouse_version_table(roundhouse_schema_name, version_table_name), ConnectionType.Default);
+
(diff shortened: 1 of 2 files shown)


RossElder1983:
Thats great news thank you.

Ill be honest im no Oracle legend myself, we came across Roundhouse as our current DB migration tool, ready roll doesnt support Oracle DBs.

But ill certainly have a look and test it out on our setup, hopefully that will be useful. We would be very interested in using the managed provider ourselves.

erikbra:
Schemas in Oracle work a bit different than e.g. in SQL server, I guess that's why the table names need to be separated by a `_` in Oracle. I don't have any good test environment for Oracle. I have looked through your changes, and they seem sensible, given that you say they work :)

I think we are breaking updating the tables if they already exist in a previous version since we are no longer using NHibernate for the migration of the RoundhousE-tables themselves, but this is a dependency I'd really like to get rid of anyway, as I don't feel the tiny usage of it is worth the dependency, so having one DB on "manual sql" is a good thing, I think. 

Could you please verify that rh off the master branch works after I have merged this PR?

PR #325, opened 2018-05-06
Added rh.bat and .zip packaging

erikbra:
It't easier to test the netcore version on Windows if distributed as a
.zip, and with at .bat file to execute.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,6 +15,7 @@ test:
       - '**\*tests.dll'
 
 after_build:
+  - 7z a .\code_drop\packages\RoundhousE.%GitVersion_SemVer%.zip .\code_drop\tmp\RoundhousE
   - 7z a .\code_drop\packages\RoundhousE.tar .\code_drop\tmp\RoundhousE
   - 7z a .\code_drop\packages\RoundhousE.%GitVersion_SemVer%.tar.gz .\code_drop\packages\RoundhousE.tar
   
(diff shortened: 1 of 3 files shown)


erikbra:
@BiggerNoise don't you think it's a good idea to include these, if we want people testing the core version on Windows too? ( I wish we could make the merge for .net core work too, but I think we're at a combination of immature tools and some ancient dependencies that make the tools not work for us yet, including log4net).

ferventcoder:
rh.bat I like, I am glad to see the zip packaging make it back in after a brief hiatus with the migration away from uppercut. One thing that is important to note is that a nuget package is simply a zip package as well.

ferventcoder:
Simply - well close to simply, it has a few more things added but it is essentially something you can rename from .nupkg to .zip and uncompress.

erikbra:
Good point about the zip vs nuget.

It is fair to remark that the .nupkg and the .zip file contains totally different bits as of now. The .nupkg contains a single, ILMerge'd rh.exe, for .NET Desktop (e.g. .NET 4.6.1), while the zip file contains the .net core version. This sadly cannot be merged into a single binary at the time, so it is a load of files. And it doesn't include Oracle or Access support (couldn't get those to work on core).

It does however run on all platforms supporting .NET Core.

So, for Windows, I think I would prefer the merged version for now if I were an end-user. But the .bat is there for the users to be able to test the .net core version on windows too.

ferventcoder:
👍

Issue #328, opened 2018-06-12
DefaultConfiguration.EnvironmentName throws exception when multiple environments are present


chuseman:
When multiple environments are specified, the new implementation of the `EnvironmentName` getter (from #297) throws:

    RoundhousE encountered an error.
    System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.InvalidOperationException: Sequence contains more than one element
       at System.Linq.Enumerable.SingleOrDefault[TSource](IEnumerable`1 source)
       --- End of inner exception stack trace ---
       at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
       at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
       at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
       at System.Reflection.RuntimePropertyInfo.GetValue(Object obj, Object[] index)
       at roundhouse.infrastructure.app.tokens.TokenReplacer.create_dictionary_from_configuration(ConfigurationPropertyHolder configuration)
       at roundhouse.infrastructure.app.tokens.TokenReplacer.replace_tokens(ConfigurationPropertyHolder configuration, String text_to_replace)
    ...

The getter uses `SingleOrDefault()`, which only allows 0 or 1 elements in the list. I'm not sure how this getter should behave when multiple environments are present -- should it just do `FirstOrDefault()` maybe?

The problem actually stems from how `TokenReplacer.create_dictionary_from_configuration()` blindly uses reflection to fill a dictionary with all of the property name/values. If instead the `ConfigurationPropertyHolder` implementation filled the dictionary, it could reason about itself and not even set an `EnvironmentName` key if there are multiple `EnvironmentNames` perhaps?

I'd be happy to contribute a PR if somebody provides some direction.

BiggerNoise:
@chuseman Thank you for reporting this. I apologize for the delay in getting back to you, it's been a madhouse here.

I agree with your idea, the configuration object should be able to take control of how it represents tokens.

So:
 - Add a to_token_dictionary() to the ConfigurationPropertyHolder interface
 - Move the implementation of create_dictionary_from_configuration to the DefaultConfiguration class renaming it so it implements to_token_dictionary()
[...]

PR #329, opened 2018-06-27
Move token dictionary creation to ConfigurationPropertyHolder

chuseman:
This will move the creation of the token dictionary out of TokenReplacer and into the ConfigurationPropertyHolder implementation. Along the way, #328 will be fixed.

A few questions:
1. The **roundhouse** project has the C# language version set to 5 - I assume this is intentional but I'd just like to make sure. nameof() and collection initialization are unavailable in C# 5.
2. I've dropped the `DefaultEncoding` and `Logger` properties from the dictionary - are there others that aren't appropriate? They're all primitive types that are at least readable as strings.
3. The `EnvironmentName` key is set to the same value as `EnvironmentNames` (comma-delimited list of environment names) - should it just be the first environment? Or only present if there is a single environment?
4. There are two implementations of `ConfigurationPropertyHolder` - is there any reason the `roundhouse.tasks.Roundhouse` implementation doesn't extend `roundhouse.consoles.DefaultConfiguration` instead of reimplementing the interface?

### DIFF
--- a/product/roundhouse.tasks/Roundhouse.cs
+++ b/product/roundhouse.tasks/Roundhouse.cs
@@ -8,6 +8,7 @@ namespace roundhouse.tasks
     using infrastructure.app;
     using infrastructure.app.logging;
     using infrastructure.containers;
+    using infrastructure.extensions;
     using infrastructure.filesystem;
     using Microsoft.Build.Framework;
     using Microsoft.Build.Utilities;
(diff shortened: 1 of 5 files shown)


BiggerNoise:
Thank you very much for this. To address your points.

1. Intentional so that we can continue to support older versions of .NET. Eventually, we'll drop that support and upgrade our version. I am quite liking many features from C#7.
2. I agree with the decision to drop those. There probably are others that shouldn't be there, but I doubt they're hurting anything.
3. I think you made the correct call here. There's no really good answer for a corner case like this.
4. I have no idea why it is done that way. I would certainly use the implementation and extend it as needed. I'd really need to dig into that code further.

BiggerNoise:
Fixes #328

PR #330, opened 2018-07-23
Add structuremap dependency to nuget package

BlythMeister:


BiggerNoise:
No. You're good. I should be able to take a look in a couple of days.
Thanks

PR #331, opened 2018-07-25
Performance Improvements (Hashing & Caching)

BlythMeister:
* Update default database to trust the cache rather than going to the database on each script check.
Based on my usage running around 2000 scripts on an empty database, this takes off around 6000 database requests
* Update the hash to be a lazy loaded string which reduces the number of times the hash is created
* Update the cached object to not include the full text of the script as it is never read.
This reduces the amount of data required to be loaded from the database

### DIFF
--- a/product/roundhouse/databases/DefaultDatabase.cs
+++ b/product/roundhouse/databases/DefaultDatabase.cs
@@ -64,7 +64,7 @@ namespace roundhouse.databases
         protected IConnection<DBCONNECTION> admin_connection;
 
         private bool disposing;
-        private Dictionary<string, ScriptsRun> scripts_cache;
+        private Dictionary<string, ScriptsRunCache> scripts_cache;
 
         //this method must set the provider
         public abstract void initialize_connections(ConfigurationPropertyHolder configuration_property_holder);
@@ -238,6 +238,7 @@ namespace roundhouse.databases
             try
             {
                 retry_policy.ExecuteAction(() => repository.save_or_update(script_run));
+                scripts_cache[script_name] = script_run;
             }
             catch (Exception ex)
             {
@@ -330,48 +331,36 @@ namespace roundhouse.databases
 
         public virtual string get_current_script_hash(string script_name)
         {
-            ScriptsRun script = get_from_script_cache(script_name) ?? get_script_run(script_name);
+            var script = get_script_run(script_name);
             return script != null ? script.text_hash : string.Empty;
         }
 
         public virtual bool has_run_script_already(string script_name)
         {
-            ScriptsRun script = get_from_script_cache(script_name) ?? get_script_run(script_name);
+            var script = get_script_run(script_name);
             return script != null;
         }
 
-        protected IList<ScriptsRun> get_all_scripts()
+        protected IList<ScriptsRunCache> get_all_scripts()
         {
-            return retry_policy.ExecuteAction(() => repository.get_all<ScriptsRun>());
+            return retry_policy.ExecuteAction(() => repository.get_all<ScriptsRunCache>());
         }
 
-        protected ScriptsRun get_script_run(string script_name)
+        protected ScriptsRunCache get_script_run(string script_name)
         {
-            QueryOver<ScriptsRun> criteria = QueryOver.Of<ScriptsRun>()
-                .Where(x => x.script_name == script_name)
-                .OrderBy(x => x.id).Desc
-                .Take(1);
+            if (scripts_cache == null)
+            {
+                scripts_cache = new Dictionary<string, ScriptsRunCache>();
 
-            ScriptsRun script = null;
-            IList<ScriptsRun> found_items;
-            try
-            {
-                found_items = retry_policy.ExecuteAction(() => repository.get_with_criteria(criteria));
-            }
-            catch (Exception ex)
-            {
-                Log.bound_to(this).log_an_error_event_containing(
-                    "{0} with provider {1} does not provide a facility for recording scripts run this time.{2}{3}",
-                    GetType(), provider, Environment.NewLine, ex.to_string());
-                throw;
+                // latest id overrides possible old one, just like in queries searching for scripts
+                foreach (var script in get_all_scripts().OrderBy(x => x.id))
+                {
+                    scripts_cache[script.script_name] = script;
+                }
             }
 
-            if (found_items != null && found_items.Count > 0)
-            {
-                script = found_items[0];
-            }
-
-            return script;
+            ScriptsRunCache script_run;
+            return scripts_cache.TryGetValue(script_name, out script_run) ? script_run : null;
         }
 
         public virtual void Dispose()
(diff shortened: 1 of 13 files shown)


BlythMeister:
Based on my build which deploys a range of database schemas here are the results.
We were on a very old (pre 2016) version of roundhouse in image 1.
The middle image is using the current nuget package version
The bottom image is using the nuget package from appveyor on this pull request.

image

Duration to deploy goes from 3 minutes 21 to 1 minute 40
Deploying the exact same databases.

BiggerNoise:
Thanks for this. As I said in your other P/R, I should be able to take a look at this in a couple of days.

BiggerNoise:
I will tell you this much. There's a bunch of gratuitous changes such as:

- include order 
- removal of comments

Please back those changes out and then squash the entire P/R into one commit. That will make it much easier to review.

BiggerNoise:
We can do a separate P/R with comment removal, but it should not be blended into another P/R

BlythMeister:
Sure, I only removed the comments as your pr checks told me too 😂

BlythMeister:
I'll rework this into a single comit with the essential changes only 👍

BiggerNoise:
Oh great, checks. Another GitHub "feature". I'll figure out how to turn that off.

General rule of thumb is that P/Rs should always be focused on one thing. 

Thanks very much for this P/R BTW, I'm really looking forward to including it.

BiggerNoise:
@BlythMeister - Did you get a message about the comments from the checks? If so, do you have that message or do you recall the details?

@ferventcoder - Can you disable the checks for this repo? I can't see them.

BlythMeister:
Updated to have less superfluous changes.

@BiggerNoise the check was this one: https://app.codacy.com/app/erikbra/roundhouse/pullRequest?prid=1948031

I wonder if it was highlighting the files i changed, and i went OTT on the cleanup...

BlythMeister:
Based on this one passing the check...I'm going to say the extra clean up was my mistake!

BiggerNoise:
Sorry, it's going to be another couple of days on this one. I picked up some kind of bug and I just can't quite focus on the code.

BlythMeister:
No worries, I've taken the app veyor build nuget package and temporarily embedded that into my repo rather than using the one on nuget.org 👍

BlythMeister:
Hey @BiggerNoise hope things are going better now, any updates on this pr?

BiggerNoise:
My year from hell has continued apace. I'm hoping that this light I see is the end of the tunnel and not another train.

I hope that I can take a closer look at this during this week.

BlythMeister:
@BiggerNoise sorry to hear this mate, hope things get better soon!

product/roundhouse/databases/DefaultDatabase.cs
```diff
@@ -238,6 +238,7 @@ public virtual void insert_script_run(string script_name, string sql_to_run, str
             try
             {
                 retry_policy.ExecuteAction(() => repository.save_or_update(script_run));
+                scripts_cache[script_name] = script_run;
```
BiggerNoise: I think that this line alone will provide 99% of the benefits of this P/R.

BlythMeister: Yeah I agree with that.  The extra hashing optimisation was a small benefit.

BiggerNoise: I'd like to have you test a more surgical version of this.
- Add the line above
- Have get_script_run() just return null (if we start with a full cache and keep it updated, we can trust the cache)
- Rerun your timing test

If that gets you more than 90% of the way there, I'd prefer to take that route.

BlythMeister: Not sure I follow?

BlythMeister: Sure I'll proceed as advised.
For me, number 2 has huge benefits because of the way that we use roundhouse and the amount of data actually stored in the roundhouse schema in the actual script column.

BlythMeister: Also, as an FYI, when I made the changes originally I did them in that order.
And 1 made about 50% of the difference
2 made about 30%
And 3 made about 20%
If I raise these changes (using lazy instead of new model) as 3 prs all extending each other with a screenshot of the difference between current and each pr, you can then made a call as to which of the 3 prs to take.
Does that sound reasonable?
I'm more that happy to put that work in, the net benefits to me and my company on this pr are huge!

BlythMeister:
@BiggerNoise Please see PRs #337 #338 & #339 & #340 for each of the changes.
Each PR adds to the previous and includes the timings of current master, any previous PRs and itself.
The timing is from my build script which calls roundhouse around 7 times as well as a bit of other logic in our deployment tool.
However all of the tests are comparable running the exact same deployment tool codebase & sql scripts against the same sql instance.
Each time this is against a fresh database as our build script does a drop prior to build.

I've already closed off #339 as this makes no difference!

I do think your right, the cache makes the biggest difference.
I'm also not sure what i have done to my scripts as the master run now as opposed to when this was originally opened has also dropped!

BlythMeister:
# Stats Summary
## Current Master
| Run1 | Run 2 | Run 3 | Average |
|---|---|---|---|
| 2:17.248 | 1:52.886 | 1:59.482 | 2:03.205 |

## Script Caching (#337)
| Run1 | Run 2 | Run 3 | Average |
|---|---|---|---|
| 1.34.084 | 1:29.291 | 1:38.848 | 1:34.074 |

## Lazy Loading Script Text (#338 & #337)
| Run1 | Run 2 | Run 3 | Average |
|---|---|---|---|
| 1:31.176 | 1:31.206 | 1:29.419 | 1:30.600 |

## New Maping (#340 & #337)
| Run1 | Run 2 | Run 3 | Average |
|---|---|---|---|
| 1.33.190 | 1:30.532 | 1:32.940 | 1:32.221 |

## Lazy hashing (#339, #338 & #337)
| Run1 | Run 2 | Run 3 | Average |
|---|---|---|---|
| 1:30.005 | 1:33.990 | 1:31.060 | 1:31.685 |

product/roundhouse/databases/DefaultDatabase.cs
BiggerNoise: Yes, that sounds like a reasonable approach. I believe that I can look at the first one on Tuesday or Wednesday.

BiggerNoise:
I have merged #337 and certainly prefer the lighter touch of #338. I assume that I wouldn't need #340 if I have #338.

I'm going to try and get #338 in tonight.

And, I really can't say it enough: thank you for the contribution, and your patience to slice and dice it so many different ways.

BiggerNoise:
#337 and #338 are merged

BlythMeister:
@BiggerNoise you are most welcome.
I can't say thank you enough for picking this project up again.
We were looking to move away from roundhouse, but since it is active again we have stayed!

Issue #327, opened 2018-06-10
Error using connectionstring


djmnz:
Tryied to upgrade from 0.8.6 to 1.0.0-RC1

I can get consistently, the cause is just attempting to enter a connection string:

`.\rh.exe /cs="Data Source=.;Initial Catalog=<DATABASE>;Persist Security Info=Tru
e;User ID=<USERNAME>;Password=<PASSWORD>"`

Got the following error:

```

Error while trying to create the BuildPlan for Instance of roundhouse.databases.Database -- Lambda: build(IContext.GetInstance(), value(roundhouse.infrastructure.app.ApplicationConfiguraton+<>c__DisplayClass3_0).configuration_property_holder).
Please check the inner exception
1.) Container.GetInstance<roundhouse.databases.Database>()

StructureMap.StructureMapBuildPlanException: Error while trying to create the BuildPlan for Instance of roundhouse.databases.Database -- Lambda: build(IContext.GetInstance(), value(roundhouse.infrastructure.app.ApplicationConfiguraton+<>c__DisplayClass3_0).configuration_property_holder).
Please check the inner exception
1.) Container.GetInstance<roundhouse.databases.Database>()
 ---> System.TypeInitializationException: The type initializer for 'StructureMap.Building.TryCatchWrapper' threw an exception. ---> System.TypeLoadException: Could not load type '<Npgsql-IChunkingTypeHandler<System-Char\[\]>-Read>d__8' from assembly 'rh, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' because the parent does not exist.
   at System.Reflection.RuntimeAssembly.GetExportedTypes(RuntimeAssembly assembly, ObjectHandleOnStack retTypes)
   at System.Reflection.RuntimeAssembly.GetExportedTypes()
   at StructureMap.Building.TryCatchWrapper..cctor()
   --- End of inner exception stack trace ---
   at StructureMap.Building.TryCatchWrapper.WrapFunc[T](String message, Type returnType, Expression expression, IDescribed described)
   at StructureMap.Building.BuildPlan.ToDelegate()
   at StructureMap.Building.BuildPlan..ctor(Type pluginType, Instance instance, IDependencySource inner, Policies policies, IEnumerable`1 interceptors)
   at StructureMap.Pipeline.Instance.buildPlan(Type pluginType, Policies policies)
   --- End of inner exception stack trace ---
[...]

PR #334, opened 2018-08-18
Replace StructureMap DI library with Ninject 

phillee007:
Replace StructureMap DI library with Ninject in order to fix issues with StructureMap exceptions when running on .Net Core. I figured this was worth it given Ninject is under active development and works in other .Net Core apps

Update tests also

Have tested locally but not yet on large project

Should fix issue #327


phillee007:
@erikbra @BiggerNoise Hopefully this can help. I've run it via rh.exe against the sample on a machine with full .net framework installed and it worked, what other testing should I do?

BiggerNoise:
I'll take a look at this at some point, but honestly, my inclination is to just get rid of the DI framework completely. I just don't think they make any sense for a small CLI application.

phillee007:
Ok thank you. I guess in the short-term without this we can't actually use RH, so need it as a way to move forward. It seems like more work to remove the DI framework than to just swap it out (which it looks like RH was already built to allow via the container abstraction)?

ferventcoder:
Definitely built with a container abstraction to make it easy to swap - I am curious why Ninject - from when I used it in the past I've found it to be the worst performing...

ferventcoder:
Sorry, posted that without the hard data - http://www.palmmedia.de/blog/2011/8/30/ioc-container-benchmark-performance-comparison. This was something started back in 2011, but the guy has a GitHub project here and has kept it up to date with results as each framework has gotten better. It's one of the things that made me choose Simple Injector back when I was looking for a replacement due to poor performance.

phillee007:
@ferventcoder because Ninject works in .Net core and has large industry usage. Performance of the DI framework didn't really factor into my thought process given the nature of RH (making db connections which are inherently slow)

erikbra:
Very happy to have more contributors, @phillee007!! I love to replace "dead" dependencies with actively developed ones. However, @BiggerNoise has a point, do we really need a DI _container_ for this project? DI, yes, but maybe we can just wire it up in the `Main()` or somewhere nearby. I'll have a quick look at how much work it would be to skip the container altogether, and get back to this PR if removing it is too big a task. I'm very keen on getting the 1.0 release out :)

erikbra:
I would prefer merging #335 in favour of this one. The DI needs are minimal in RoundhousE, and removing a dependency altogether instead of replacing it with another is better, ihmo. Any opinions?

BiggerNoise:
@phillee007 - Thank you for the contribution, even if we decided to head a different direction.

phillee007:
No worries - I'm glad it got resolved either way.

PR #335, opened 2018-09-05
Remove dependency on StructureMap

erikbra:
The current DI pattern in RoundhousE is somewhat Service Locator based, so coding away the concept of a global, static `Containter` class is a bit more of a job. IMHO, dependencies should be set up and injected into the consuming classes/components from one "starting point" in the program (typically `Main()` in a console app), and not registered in a single, static container. But this is probably a discussing point.

Without changing the overall architecture, since we don't use any of the more advanced feature of a DI container (pooling, disposing instances, etc), I have replaced StructureMap with a single hard-coded "Container" that serves the purpose of a backend to how we use the static `Container.Resolve` in RoundhousE.

It's no rocket science, but since we had problems with StructureMap, and StructureMap is abandoned, we might as well get rid of the dependency. The fewer dependencies, the better, as long as we don't have to invent very advanced wheels ourselves, IMHO.

This solves #327 (hopefully).


erikbra:
Could you please take a look at this when you have a chance, @BiggerNoise?

erikbra:
The codacy hint it bullsh*t, btw, this is an implementation of an interface. I can't make it static...
https://app.codacy.com/app/ferventcoder/roundhouse/pullRequest?prid=2146165

BiggerNoise:
Also agree that the service locator approach should be replaced, but this is an excellent step forward

Issue #332, opened 2018-07-29
Publish RoundhousE as a dotnet global tool


erikbra:
Literature:

https://natemcmaster.com/blog/2018/05/12/dotnet-global-tools/
https://github.com/natemcmaster/dotnet-globaltool-templates

Example of other global tool:

https://github.com/jerriep/dotnet-outdated

- This is dependent on the 1.0.0 release, _maybe_ it should be part of it? 
- Requires upgrade from .net core 2.0 to 2.1 (2.0 is not a long-term support release anyway=

Would enable being able to install RoundhousE (the .net core version - including multi-platform support, excluding some providers) with the following command, given that .net core is installed:

`dotnet tool install -g roundhouse`

And that's it...

Don't think this would be too much work.

BiggerNoise:
I don't think the 2.1 is an issue at all.

If this isn't 1.0, then it should be in 1.0.1. I'm not sure we're going to have an ILMerge for core stuff, this seems like a somewhat better approach.

erikbra:
Solved in PR #336

BiggerNoise:
Merged

mattiasnordqvist:
is this working, really?

PR #336, opened 2018-09-07
Issue 332 - Publish as a dotnet core global tool

erikbra:
This adds publishing of RoundhousE as a .net core global tool, which drastically simplifies installation of the .net core version of RoundhousE on all platforms.

To test the installation, before the tool is actually pushed to nuget, you need to add the nuget folder on install, like this (from the project root folder):

1. `./build.ps1`
1. `dotnet tool install --add-source ./code_drop/packages --version 1.0.0-RC.2 -g dotnet-roundhouse`

(change version number as needed)

You can then see the tool is installed:

`dotnet tool list -g`

output from my box:
```
Package Id             Version         Commands       
------------------------------------------------------
dotnet-outdated        2.0.0           dotnet-outdated
dotnet-roundhouse      1.0.0-rc.2      dotnet-rh      
```

And you can uninstall it:

`dotnet tool uninstall -g dotnet-roundhouse`

Solves #332 .

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -51,26 +51,24 @@ Chocolatey is like apt-get, but for Windows! This is an alternative method to ge
 2. Then from anywhere you can type `rh [options]`  
 
 ### Dotnet core (Linux, macOS, etc)
-We have a new build of RoundhousE, built on the new(er) mean and lean dotnet core. Dotnet core applications are a bit different that standard
-.net desktop (e.g. .Net 4.6.1) applications, and we're still figuring out the best way to distribute them. If you have opinions on what
-is the most convenient form for a particular platform, please let us know. Even better, if you have knowledge of a particular distribution method,
-or would like to learn about one, file a Pull Request with changes in the build script to 
-produce what you want.
+We have a new build of RoundhousE, built on the new(er) mean and lean dotnet core. 
+Dotnet core runs on all major OS-es, including Windows, Linux and macOS.
+RoundhousE is distributed as a .net core 2.1 _global tool_, and can be installed like this:
 
-For now, we publish a `.tar.gz` of the application, at the [Releases](https://github.com/chucknorris/roundhouse/releases) page. Extract the file somewhere, and run 
-the included `rh` shell script. You will need dotnet core installed on your box for this to work. You can get it here: [https://dot.net](https://dot.net).
+`dotnet tool install -g dotnet-roundhouse`
 
-So: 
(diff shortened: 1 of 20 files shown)


BiggerNoise:
I'm having trouble getting the install to run:
```
PS C:\development\Chuck\roundhousE\master-1.0> dotnet tool install --add-source ./code_drop/ --version 1.0.0-RC.2 -g dotnet-roundhouse
The settings file in the tool's NuGet package is invalid: Entry point file 'rh.dll' for command 'dotnet-rh' was not found in the package.
Tool 'dotnet-roundhouse' failed to install. Please contact the tool author for assistance.
```

Probably something very simple, but dotnet tools are pretty new to me as well.

I did:
- Install Code SDK 2.1
- Download your branch
- build.ps1
- Attempt to run the command.

Let me know if you have a clue you can share.

BTW - I assume that the real publishing takes place when we push the dotnet-roundhouse.xxx package to NuGet, right?

erikbra:
@BiggerNoise thanks for taking the time to look at this. I had two errors - one in the `.csproj`, making it fail to build on windows (RuntimeIdentifier stuff), and one doc error. I had moved the drop location to `code_drop\packages`, together with the other nuget packages. I have corrected the path in the nuget add source command above, sorry. 

So the correct command line would be:

`dotnet tool install --add-source ./code_drop/packages --version 1.0.0-RC.2 -g dotnet-roundhouse`

Please try again when you have a chance. 

And, yes, this is only available to people without first building it themselves and adding a local nuget repo after we push it to nuget. btw, are we happy with the package name? The other ones are called

`roundhouse.*`

Maybe we should call it `roundhouse.dotnet-tool` or something, althought it looks a bit strange to me. Keeping the roundhouse prefix could be a good thing. I don't know if there's a standard on how to name dotnet core global tools' packages on nuget.

BiggerNoise:
Awesome addition.
I'm not crazy about the current name, but don't have a definitive suggestion for an alternate.
Perhaps "roundhouse" I don't think we need the dotnet spelled out. If we feel we had to have it, we could abbreviate as "dn".

- roundhouse
- rh-dn
- rh-tool

@davidboothe - You want to give this a try on your Mac?

erikbra:
Thanks for the merge, @BiggerNoise. You're really on a roll RoundhousE-wise these days. Lots of PRs getting merged!

Re the naming, I agree. There are two ways of naming/running `dotnet` global tools, one is the `dotnet commandname`, the other just `commandname`. One chooses when building the package which form to use, if I have understood it correctly. See e.g.

https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools#use-the-tool

We could have named the tool just `roundhouse`, but the disadvantage of that is that I think the users need to put the install location in their path for it to work.

PR #353, opened 2018-12-11
Issue 352 - Can't find Antlr

erikbra:
Run dotnet publish before pack, to get all dependencies in the package.


erikbra:
Was able to install this on my local box, off the build server. Should work, then.

PR #357, opened 2018-12-18
Fixes Issue 349: Problems Parsing MySQL Delimiter

cmiles74:
I found some issues with the MySQL statement parser that were causing problems. In addition to the issues with trailing delimiters, I also saw some problems parsing parenthesis in views that had many joins. Typically this was from DDL produced by MySQL itself (i.e., the DDL was pulled from MySQL, edited, and then RoundHouse was executing the edit).

I re-wrote the parsing code with an eye specifically towards the minimum amount of parsing that would be required to render reliable statements that we could then execute. This parser will parse and handle the following types of statements:

* Statements that define the delimiter
* Statements of SQL

To support that I coded up a scanner that handles the minimum tokens we need to let us parse out those statements. The scanner will produce the following types of tokens:

* Delimiter declaration keyword
* Delimiters
* Comments (delimiters in comments are ignored, comments in stored procedured are retained)
* Quote (delimiters in quotes are ignored)
* ANSI Quote (delimiters in ANSI quotes are ignored if ANSI quotes are being honored)
* White space (we want to retain the white space present in stored procedures)
* Text (delimiters in text denote the end of the statement)

I have tested this with my own SQL and it works well. It handles nested parenthesis as well as trailing delimiter declarations.


erikbra:
This is great stuff, @cmiles74! Love that you have separated out the parser logic into a separate class. There are a couple of linter errors. There is a `roundhouse.sln.DotSettings` file in the root folder that controls these rules. They may be a bit unfamiliar (at least they were to me). I don't know if you have access to e.g. Visual Studio or Jetbrains' Rider IDE, I think both of these can give you the linter tips directly in the IDE. I don't think there is any plugin support for `.DotSettings` files in VSCode, which you seem to be using.

Other than this, stuff looks good. What would make it even better, though, is unit tests. I would really appreciate if you could make some simple tests that demonstrate the logic in the parser. I know the test coverage of the parsing is virtually non-existant for MySql, but, especially since I'm not a MySql pro, tests would really help me understand and validate the logic of the parser.

That said, splitting out into parser classes is definitely a step in the right direction, and I think it this in time could be generalised to be used in the other DB providers too.

product/roundhouse.databases.mysql/parser/Parser.cs
```diff
@@ -0,0 +1,160 @@
...
+        {
+            bool value = false;
+
+            if (current >= tokens.Count) {
+                value = true;
+            }
+
+            return value;
+        }
+    }
```
erikbra: This is nitpicking, I'm sorry, but I really would simplify this to

```suggestion
        private bool IsAtEnd()
        {
            return current >= tokens.Count;
        }
```

product/roundhouse.databases.mysql/parser/Parser.cs
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        private ParsedStatement ParseStatement() 
+        {
+            Token token = Advance();
+            string delimiter = ";";
+            bool setDelimiter = false;
+            ParsedStatement.Type statementType = ParsedStatement.Type.Sql;
+
+            while (!IsAtEnd() && !token.TokenType.Equals(Token.Type.Delimiter)) {
```
erikbra: I like the style. Can be read as a sentence :)

product/roundhouse.databases.mysql/parser/Parser.cs
```diff
@@ -0,0 +1,160 @@
...
+            return statements;
+        }
+
+        private ParsedStatement ParseStatement() 
+        {
+            Token token = Advance();
+            string delimiter = ";";
+            bool setDelimiter = false;
+            ParsedStatement.Type statementType = ParsedStatement.Type.Sql;
+
```
erikbra: I like the style. Can be read as a sentence :)

product/roundhouse.databases.mysql/parser/Parser.cs
```diff
@@ -0,0 +1,160 @@
...
+namespace roundhouse.databases.mysql.parser
+{
+
+    /// <summary>
+    /// Provides a parser that will parse statements out of a string of MySQL SQL.
+    /// </summary>
+    public class Parser
+    {
+
+        /// <summary>
```
erikbra: + for doc comments :)

cmiles74:
I think this last commit covers all of the remaining static analysis issues. There are five items still flagged, but they seem more like a matter of style. Let me know if you'd like them changed one way or the other. 

Thank you! :-)

erikbra:
I'm really fine with the remaining linter issues, no worries.

But I would really appreciate some tests on the new classes, at least the parser, ideally with some sample SQL (inline or in files, I don't care), and test on the desired splitting behaviour. Would you be able to create some as part of this PR?

cmiles74:
Understood, I will try to get some together in the next day or two.

erikbra:
That would be perfect, thanks. And, again, I really appreciate the effort and attention to quality and detail of this PR.

product/roundhouse.tests/sqlsplitters/MySqlStatementSplitterSpecs.cs
```diff
@@ -0,0 +1,152 @@
...
+            {
+                string script = "/*\n *\n * Test Statement\n *\n */\nselect * from test1;\nselect * from test2;\n";
+
+                Parser parser = new Parser(script);
+                List<ParsedStatement> statements = parser.Parse();
+                WriteStatements(statements);
+
+                TestContext.Out.WriteLine("Statements parsed: " + statements.Count);
+                Assert.IsTrue(statements[0].Value.Length > 25);
+                Assert.AreEqual(statements.Count, 2);
```
erikbra: You've got the expected and actual mixed up here (all of them are). It should be
```suggestion
                Assert.AreEqual(2, statements.Count);
```

product/roundhouse.tests/sqlsplitters/MySqlStatementSplitterSpecs.cs
```diff
@@ -0,0 +1,152 @@
...
+            public void should_split_on_delimiters()
+            {
+                string script = "select * from test1;\nselect * from test2;\n";
+
+                Parser parser = new Parser(script);
+                List<ParsedStatement> statements = parser.Parse();
+                WriteStatements(statements);
+
+                TestContext.Out.WriteLine("Statements parsed: " + statements.Count);
+                Assert.IsTrue(statements[0].Value.Length > 10);
```
erikbra: Nitpicking again, sorry. And, I do realise that the readability of the tests in RoundhousE is not great, but we always try to improve, don't we? ;), but I would prefer if you used the `Assert.Greater` instead, as the output if the test fails is much better: 

```suggestion
Assert.Greater(statements[0].Value.Length, 10);
```

Example, see if I change the test to test on 50 instead of 10. The `Assert.IsTrue` outputs:
```
roundhouse.tests.sqlsplitters.MySqlStatementSplitterSpecs+when_splitting_sql_statements.should_split_on_delimiters

  Expected: True
  But was:  False
```

However, if you use `Assert.Greater`, it outputs exactly what is the problem, so fixing the error is much easier:

```
roundhouse.tests.sqlsplitters.MySqlStatementSplitterSpecs+when_splitting_sql_statements.should_split_on_delimiters

  Expected: greater than 50
  But was:  20
```

This is _not_ meant as picking on you, definitely, just hoping to improve the quality of the tests of RoundhousE going forward.

Another thought, what does this assert really test? Yes, it tests that the first token is longer than 10, but it that the essence of the test? I see that testing that the count is 2 is definitely testing the logic, and makes sense. I just wonder what scenario the first assert is testing.

*Please*, if you feel this it tedious, and I'm being a paint in the butt, do tell me. It's well meaning, but if you feel it's a bit too much, just say so, and I'll be happy to merge the PR. It does add a lot of business value (i.e. fixing a bug) :)

product/roundhouse.tests/sqlsplitters/MySqlStatementSplitterSpecs.cs
```diff
@@ -0,0 +1,152 @@
...
+            public void should_retain_comments()
+            {
+                string script = "/*\n *\n * Test Statement\n *\n */\nselect * from test1;\nselect * from test2;\n";
+
+                Parser parser = new Parser(script);
+                List<ParsedStatement> statements = parser.Parse();
+                WriteStatements(statements);
+
+                TestContext.Out.WriteLine("Statements parsed: " + statements.Count);
+                Assert.IsTrue(statements[0].Value.Length > 25);
```
erikbra: A suggestion, test that the first statement actually contains the comment string, instead of testing on the length:

```suggestion
                 Assert.That(statements[0].Value.Contains("Test Statement"));
```

product/roundhouse.tests/sqlsplitters/MySqlStatementSplitterSpecs.cs
```diff
@@ -0,0 +1,152 @@
...
+            public void should_honor_tricky_single_character_delimiter()
+            {
+                string script = "delimiter Z\nselect * from test1Z\nselect * from test2Z\n";
+
+                Parser parser = new Parser(script);
+                List<ParsedStatement> statements = parser.Parse();
+                WriteStatements(statements);
+
+                TestContext.Out.WriteLine("Statements parsed: " + statements.Count);
+                Assert.IsTrue(statements[0].Value.Length > 10);
```
erikbra: Same here - I see that the resulting statement contains a trailing `\n`, even thought the delimiter `$` is before the `\n` in the original string. Is that expected? It would be great if the test expected the exact expected outcome, as it is much easier to understand the intent of the parsing - e.g. if the trailing newline should be included in the statement or not.

Could you please change the assertion accordingly, i.e:
 Assert.That(statements[0].Value.Contains("Test Statement"));
```suggestion
                Assert.AreEqual(statements[0].Value, "select * from test1\n");
```

(the above fails as of now)

or, if you didn't mean to have the newline in the statement,

 Assert.That(statements[0].Value.Contains("Test Statement"));
```suggestion
                Assert.AreEqual(statements[0].Value, "select * from test1");
```

product/roundhouse.tests/sqlsplitters/MySqlStatementSplitterSpecs.cs
cmiles74: Okay, I will get these straightened out.

product/roundhouse.tests/sqlsplitters/MySqlStatementSplitterSpecs.cs
cmiles74: My concern is that it would do something weird, like only count the trailing delimiter and provide two statements: one with all of the test text and a second statement that was empty or only white space.

Another thing I was trying to test was the preservation of white space and comments. Most of the time these don't matter but I've seen people put a lot of comments in stored procedures.

I will clean these up a bit and try to make the things being tested a little more clear.

Thank you!

product/roundhouse.tests/sqlsplitters/MySqlStatementSplitterSpecs.cs
cmiles74: I changed this to contain the expected string, it did expose a bug with the multi-line comment handling. This is working now and the test passes.

product/roundhouse.tests/sqlsplitters/MySqlStatementSplitterSpecs.cs
cmiles74: Each parsed statement is the text of the statement, minus the delimiter. For MySQL, anyway, it doesn't want to receive the actual delimiter and treats every statement we send as a complete statement.

When tokens are gathered into statements, every "EndOfLine" token results in the correct line ending for the environment being appended (Environment.NewLine). In this case the "$" is being left off (the delimiter isn't being sent) and then, I believe, the trailing line ending is being added because the next token is an EndOfLine.

I think I've changed all of these test cases to be more explicit.

Thank you!

PR #362, opened 2019-01-26
Add docker distribution of RoundhousE

### LINKED ISSUE #342: Distribute RoundhousE on Docker?
Hi all,

Would it be possible to officially distribute a version of RoundhousE in a .Net core based docker container? I've whipped together a quick version of RoundhousE running in a docker container, but it's probably not the best way to do it:

https://github.com/namehillsoftware/dotnet-core-roundhouse/blob/master/Dockerfile

erikbra:
Added Dockerfile, which installs `dotnet-roundhouse` on a Microsoft 2.2. sdk .net core sdk image.

Fixes #342

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM microsoft/dotnet:2.2-sdk
 
-MAINTAINER erik@brandstadmoen.net
+LABEL maintainer="erik@brandstadmoen.net"
 
 ENV PATH="$PATH:/root/.dotnet/tools"
 RUN dotnet tool install dotnet-roundhouse -g
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM microsoft/dotnet:2.2-sdk
+
+MAINTAINER erik@brandstadmoen.net
+
+ENV PATH="$PATH:/root/.dotnet/tools"
+RUN dotnet tool install dotnet-roundhouse -g
+
+ENTRYPOINT [ "rh" ]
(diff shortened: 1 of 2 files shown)


erikbra:
@danrien, would you mind taking a look at this PR? IT uses a Dockerfile very similar to the one you linked to in #342 . I am also thinking of the documentation - do you think you could provide some usage scenarios using the docker image? E.g how it's used in a pipeline, etc, and that it generally makes sense. It would be great if you could.

erikbra:
Fixes #342

danrien:
Hey @erikbra, sorry I didn't get a chance to get back to you on this sooner. The only suggestion I would have made would have been to tag the image - https://docs.docker.com/engine/reference/commandline/tag/, and likewise, to pin the version of RoundhousE in the image. Microsoft's images are a good reference to follow IMO:

Dotnet image: https://hub.docker.com/r/microsoft/dotnet/
Source: https://github.com/dotnet/dotnet-docker/blob/master/2.2/sdk/stretch/amd64/Dockerfile

erikbra:
Sorry, I merged it a bit early, I didn't expect you to reply so soon, but I was running a release, and thought I might include the merge, when I was already on it. 

I'm not sure what you mean by tagging the image. I have tagged it here, if that is what you mean? Or do I misunderstand? https://hub.docker.com/r/dotnetroundhouse/roundhouse/tags

cpoDesign:
@erikbra Eric while there is new functionality added, is it possible that the wiki page is added also? I love the idea of using RoundhouseE from a container, but there is no documentation for it.

PR #364, opened 2019-02-01
Issue 363

erikbra:
More and more dependencies have issues with being run from inside an ILMerged assembly. Stumbled across https://github.com/Fody/Costura, which embeds dependencies in the assembly, and uses Code weaving to add assembly loading hook to load from embedded resource, instead of merging the assemblies in the traditional way.

It looks promising so far.


erikbra:
Solves issue #363

Issue #368, opened 2019-02-07
ReliableSqlConnection holds connections open


BlythMeister:
The ReliableSqlConnection will never close the connection...even when it should.

BlythMeister:
i am working on a PR to fix this.

BlythMeister:
This was my bad, nothing to see here....

erikbra:
OK, so you actually meant to abandon the PR?

BlythMeister:
Yeah, I misread the code...it was something else holding connection in my code 🤣

erikbra:
It happens. A lot. To everyone ;)

PR #369, opened 2019-02-07
Fix for #368 - Close connection in a finally

BlythMeister:
This ensures that a connection is actually closed when it should be


erikbra:
Did you mean to close this PR, @BlythMeister , don't you want it merged?

PR #371, opened 2019-02-10
Version roundhouse in the docker image

danrien:
Hi @erikbra,

I thought I might just create a PR with some suggestions for improvements on the docker image definition. Of note are the following changes:

- Add an explicit version to `RUN dotnet tool install dotnet-roundhouse -g --version $roundhouse_version` - this is the most important change. It pins the version of RoundhousE in the image so that consumers of the image definition can depend on the right version of RoundhousE being present in their containers
- Pass in the version during image build time with an `ARG` command
- Build the image in `build.ps1`, and create a "latest" and a versioned tag for the image
- Add some AppVeyor lines that might push the final images to DockerHub?

Unfortunately, since I normally develop on a Linux host at home, I can't easily test out the update to the build script without a reboot into Windows. Normally, I don't open a PR without some testing of my own, but the amount of work involved in getting my Windows environment into a build-ready environment shied me away from waiting that long to open this PR.

Beyond testing the build, I'm not sure how I could test out the AppVeyor changes. I am concerned that the changes could break other pieces of the AppVeyor build, but that mostly stems from complete inexperience with AppVeyor! Also, a couple of environment variables would need to be set (by you or another maintainer, I'd presume) in order for the push to DockerHub to work.

Thanks!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,9 @@ FROM microsoft/dotnet:2.2-sdk
 
 LABEL maintainer="erik@brandstadmoen.net"
 
+ARG roundhouse_version
+
 ENV PATH="$PATH:/root/.dotnet/tools"
-RUN dotnet tool install dotnet-roundhouse -g --version 1.0.4
+RUN dotnet tool install dotnet-roundhouse -g --version $roundhouse_version
 
 ENTRYPOINT [ "rh" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,6 @@ FROM microsoft/dotnet:2.2-sdk
 LABEL maintainer="erik@brandstadmoen.net"
 
 ENV PATH="$PATH:/root/.dotnet/tools"
-RUN dotnet tool install dotnet-roundhouse -g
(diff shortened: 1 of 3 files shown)


erikbra:
Hi, @danrien - terribly sorry for the ridiculously long feedback time from my part here. Thank you for your contribution. It looks good to me, just a quick check, this would make the docker image version match the roundhouse version, am I understanding correctly?

As to testing stuff on localhost, there's no problem running Powershell scripts on Linux. Just install powershell core, and you'll be able to run poweshell on Linux too (https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell-core-on-linux?view=powershell-6).

I'll look into deploying directly to docker hub from AppVeyor. We only push to MyGet from AppVeyor ATM, nuget, chocolatey, and docker are manual. I'd at least like it to only push stable builds to docker hub, this build script chang wuild push every build, would it not?

erikbra:
This one is not needed anymore, as we don't run the .net core global tool in the docker image anymore, just compile and run RoundhousE ourselves.

PR #376, opened 2019-03-15
Updated parser to handle single quoted values

cmiles74:
I ran into an issue this week and it turns out (much to my embarrassment) that the MySQL parser doesn't handle single (') quoted values, only back-tick or double quoted. This change adds single quoted values to the parser. It also adds a new test case for single quoted values.

I have another project where I test these changes against a database, you can find a [more practical example there](https://github.com/cmiles74/rh-mysql-sproc-test/blob/master/up/0003_Parsed_Statement.sql).

Thank you! `:-D`


cmiles74:
If there's anything I can do to move this pull request closer to acceptance, let me know. :-D

PR #385, opened 2019-09-20
Feature/redshift support

neoaisac:
Adds initial AWS Redshift support for RoundhousE.

### DIFF
--- a/product/roundhouse.core/infrastructure.app/DatabaseTypeSynonyms.cs
+++ b/product/roundhouse.core/infrastructure.app/DatabaseTypeSynonyms.cs
@@ -54,6 +54,9 @@ namespace roundhouse.infrastructure.app
                 case "postgresql":
                     database_type_full_name = "roundhouse.databases.postgresql.PostgreSQLDatabase, roundhouse.databases.postgresql";
                     break;
+                case "redshift":
+                    database_type_full_name = "roundhouse.databases.postgresql.RedshiftSQLDatabase, roundhouse.databases.postgresql";
+                    break;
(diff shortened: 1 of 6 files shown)


neoaisac:
@erikbra I thought I'd do the PR as we're already running locally with an alpha version for Redshift based on this code. I'd like to have someone reviewing the code (it's my first public OSS commit in years) :)

erikbra:
@neoaisac Thanks a lot for your PR. And terribly sorry for the long time before any feedback from my part. Work and other parts of my life has made RoundhousE bump down the list of priorities. I think your PR looks nice and clean.

Just a quick question, I don't know Amazon Redshift, and I just wondered why you have made the implementations subclass NpgSql? Is Redshift based on PostgreSQL?

neoaisac:
Hi @erikbra. Redshift uses a somewhat compatible interface based on PostgreSQL 8.0.2, however it's not 100% PostgreSQL. There are some things that don't work, and some things that don't work as expected in PostgreSQL standard.

Maybe this helps:

- https://docs.aws.amazon.com/redshift/latest/dg/c_redshift-and-postgres-sql.html
- http://www.npgsql.org/doc/compatibility.html

erikbra:
Thanks a lot for the update, @neoaisac . I tried reading up on it on the Amazon Redshift website, but my google fu wasn't up to par, apparently :)

I presumed it was based upon PostgreSQL, since you made it inherit that, I just wanted to be sure and know a bit more before merging. 

Again, thanks a lot for your PR.

PR #388, opened 2019-10-10
Fix EOL-related hash comparison bug

jjanuszkiewicz:
Hash comparison should ignore end-of-line variations, but it fails in one scenario:
 - when the original hash was calculated by an older RH version, which didn't normalize the script to LF line endings before hashing
 - and the original script had CRLF line endings
 - and the same file, but with LF line endings, is executed by the migrator.

This PR refactors hash generation and comparison into a separate class, adds unit tests that demonstrate the bug and fixes it.


erikbra:
@jjanuszkiewicz  Nice separation of the hash algorithm into a separate class and interface. And good catch. Thank you very much for your contribution! :)

jjanuszkiewicz:
@erikbra Thanks for merging this :-) Any idea when 1.2.0 will be released? My team is waiting for this fix, as we've actually been hit by this problem in practice. I'd rather not build our own temporary clone of RH if I can avoid it.

erikbra:
The plan is to get it out by the end of the month. I have two more features I would like to include (open PRs), if possible. https://github.com/chucknorris/roundhouse/milestone/10

Issue #354, opened 2018-12-11
Is it possible to remove RoundHousE_ScriptsRun text_of_script column?


ido-ran:
I'm working on a project that uses RoundHousE migrations. We accumulated lots of migrations over a 3 years period.
I've done an analysis on our database and found that RoundHousE_ScriptsRun is our 2nd largest table and it takes 72MB on disk. When I removed `text_of_script` column the table dropped to 160KB on disk.

Is it possible to remove that column?

We don't need the full script text in the database. We have the files in source control.

erikbra:
Of course, this could be an option. It depends on what people want. I don't use the RoundhousE tables much myself, for selects and stuff, but other people's mileage may vary. I think removing the column could break others' use of the tables, but, technically, only the hash is needed for RoundhousE to work correctly, if I'm not mistaken.

Anyone else out there with opinions? I won't go ahead and remove the column before gathering mor opinions. I see your concern, but people might be using it. We could, of course, make it an option. But, ironically enough, versioning of the Roundhouse.* tables are a weak point in RoundhousE, so changes are not so easily made either.

That said, 72MB isn't exactly a huge database table. People have tens and hundreds of GBs in their databases, but, of course, people have different sizes and different applications.

ido-ran:
Any update on this issue?

erikbra:
Sorry for the long lead-time. PR #389 should solve this once resolved.

PR #389, opened 2019-10-12
Add switch to not store full text of scripts run

peckham:
Provides a solution for #354 by adding a new configuration property

### DIFF
--- a/product/roundhouse.console/Program.cs
+++ b/product/roundhouse.console/Program.cs
@@ -359,6 +359,10 @@ namespace roundhouse.console
                 .Add("isuptodate",
                      "This option prints whether there are any database updates or not, without actually running them. Other output except errors is disabled, to make it easy to use in scripts.",
                      option => { })
+                //do not store full text of run scripts
+                .Add("donotstorescriptsruntext",
+                     "DoNotStoreScriptsRunText - This instructs RH to not store the full script text in the database. Defaults to false.",
+                     option => configuration.DoNotStoreScriptsRunText = option != null)
                 // default encoding
                 .Add("defaultencoding=",
                      "Default encoding to use for loading script file from disk if file doesn't contain BOM. For the list of possible values see the column Name in table listed in .NET Encoding class documentation. Defaults to UTF-8",
--- a/product/roundhouse.core/migrators/DefaultDatabaseMigrator.cs
+++ b/product/roundhouse.core/migrators/DefaultDatabaseMigrator.cs
@@ -256,7 +256,7 @@ namespace roundhouse.migrators
         public void record_script_in_scripts_run_table(string script_name, string sql_to_run, bool run_this_script_once, long version_id)
         {
             var hash = hash_generator.create_hash(sql_to_run, true);
-            if (!configuration.DoNotStoreScriptsRunText)
+            if (configuration.DoNotStoreScriptsRunText)
             {
                 sql_to_run = null;
             }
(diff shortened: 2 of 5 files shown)


product/roundhouse.core/migrators/DefaultDatabaseMigrator.cs
```diff
@@ -255,12 +255,23 @@ public IEnumerable<string> get_statements_to_run(string sql_to_run)
 
         public void record_script_in_scripts_run_table(string script_name, string sql_to_run, bool run_this_script_once, long version_id)
         {
+            var hash = hash_generator.create_hash(sql_to_run, true);
+            if (!configuration.DoNotStoreScriptsRunText)
```
erikbra: Isn't this one inverted? Shouldn't the `sql_to_run` be set to `null` if `DoNotStoreScriptsRunText` is set to `true`?

product/roundhouse.core/migrators/DefaultDatabaseMigrator.cs
```diff
@@ -255,12 +255,23 @@ public IEnumerable<string> get_statements_to_run(string sql_to_run)
...
+            }
+
             Log.bound_to(this).log_a_debug_event_containing("Recording {0} script ran on {1} - {2}.", script_name, database.server_name, database.database_name);
-            database.insert_script_run(script_name, sql_to_run, hash_generator.create_hash(sql_to_run, true), run_this_script_once, version_id);
+            database.insert_script_run(script_name, sql_to_run, hash, run_this_script_once, version_id);
         }
 
         public void record_script_in_scripts_run_errors_table(string script_name, string sql_to_run, string sql_erroneous_part, string error_message, string repository_version, string repository_path)
         {
+            if (!configuration.DoNotStoreScriptsRunText)
```
erikbra: Same comment. I think this one  should not be negated?

product/roundhouse.console/Program.cs
```diff
@@ -359,6 +359,10 @@ private static void parse_arguments_and_set_up_configuration(ConfigurationProper
                 .Add("isuptodate",
                      "This option prints whether there are any database updates or not, without actually running them. Other output except errors is disabled, to make it easy to use in scripts.",
                      option => { })
+                //do not store full text of run scripts
+                .Add("donotstorescriptsruntext",
+                     "DoNotStoreScriptsRunText - This instructs RH to not store the full script text in the database. Defaults to false.",
+                     option => configuration.DoNotStoreScriptsRunText = option != null)
```
erikbra: Is it better to name it "Positive"? i.e. `StoreScriptsRunText`, and default it to `true`? I think double negations are a bit hard to relate to (see also my comments further down, I think there is a negation too much in the code).

See e.g. https://softwareengineering.stackexchange.com/a/196837 for a discussion on affirmative (isXXX) vs negative (IsNotXXX) names on boolean variables.

Feel free to disagree, by all means :)

peckham: I do agree on the affirmative naming. That's what I had initially (hence the forgotten negations), but it didn't quite feel consistent with `DoNotCreateDatabase` and `DoNotAlterDatabase`. That's why I changed it to the negative name.

I'll try to update my PR tomorrow :)

erikbra: Fair point. Maybe we should keep it as "DoNot". There are arguments for making the "unordinary" run cases a "true" flag too, if you see what I mean. So I may be convinced in both directions, really. So, keep the negative form if you like. But, please have a look at the concerns with the double negations of the flag in the code. I think the logic is reversed (but I might be wrong)

peckham: You're right about the double negations. I've fixed them in the latest commit

erikbra: Thanks. Tested it locally, seems to work great. Thanks a lot for your contribution!!

Issue #275, opened 2017-08-24
-dryrun still writing a version record to [RoundhousE].[Version] table


JackGraber:
Running the 0.8.7 code and testing out the -dryrun switch.  It indeed kept scripts from running against the target database.  But it wrote a new version record in the [RoundhousE].[Version] table every time. 

Tested against SQL Server 16 and Azure.

erikbra:
Fixed in #392

PR #392, opened 2019-10-14
Prevent dryrun from writing version records

peckham:
Provides a fix for #275 
If unit tests are to be written, please guide me in the right direction for the correct class.


erikbra:
Lovely. Thanks a lot! To be honest, I am not very familiarized with the way the tests are written in this project. They are very BDD-style, and I'm not very experienced with that style. I often get confused. I have read through your code, and it looks good. Tests would be nice, but I don't have a good advice on where to write them :)

Issue #386, opened 2019-09-26
RH on linux running up scripts in "reverse" order.


coactivechris:
This one took a while for me to work out, and I hope someone can help with how best to get around it.

So we have a folder structure for our up scripts:
```
- up
  - 0000_FirstScripts
    - 0001-Script1.sql
    - 0002-Script2.sql
  - 0001_SecondScripts
    - 0001-Script3.sql
    - 0002-Script4.sql
 - 0002_ThirdScripts
    - 0001-Script5.sql
    - 0002-Script6.sql
```

etc.

Now, when running this via the RH dotnet tool on windows, scripts are run in the expected order, top to bottom, from 0000 directory on downwards.  

When running the same using the RH dotnet tool on linux, the order of the scripts run appears to be reversed, that is, starting with the LAST folder, 0002-ThirdScripts, then running the scripts inside that folder in top to bottom order, e.g. 0001-Script5.sql, 0002-Script6.sql, then back to 0001-SecondScripts, etc.

Now , our naming convention might be a bad one, so what naming standards are recommended to ensure things are run in the right order.  In reading the source code, I see it is reading the filename from a path only, without the folder.  Would we need to name scripts as all numbered away from one another, irrespective of subdirectory?

Let me know

JscNZ:
Should be resolved once #393 is released 👍

PR #393, opened 2019-10-15
Order directories by path

JscNZ:
As seen in #386 directories are not ordered when they're retrieved. This results in different behaviour when executing up scripts on windows vs linux (assuming it's actually different due to the way the different filesystems work under the hood).

This adds an option to force the directories to be retrieved in order so that scripts run in a repeatable manner. The config defaults to false to avoid breaking any existing executions based on the old behaviour.

image
Running `Directory.GetDirectories(upPath)` on windows vs linux

### DIFF
--- a/product/roundhouse.core/infrastructure/filesystem/DotNetFileSystemAccess.cs
+++ b/product/roundhouse.core/infrastructure/filesystem/DotNetFileSystemAccess.cs
@@ -370,7 +370,7 @@ namespace roundhouse.infrastructure.filesystem
         /// <returns>A list of subdirectories inside of the existing directory</returns>
         public string[] get_all_directory_name_strings_in(string directory)
         {
-            return Directory.GetDirectories(directory);
+            return Directory.GetDirectories(directory).OrderBy(d => d).ToArray();
         }
 
         /// <summary>


erikbra:
@JscNZ  Thanks a lot for you contribution. Just to be clear, `0010_Add_orderedParts`, `0015-PRIC-154`, etc, are _directories_ in the `up` folder, not files?

I think maybe the configuration flag is not necessary. Even though not explicitly stated that directories should be read in alphabetical order, I think the documentation here implies that directories should also be read alphabetically.

https://github.com/chucknorris/roundhouse/wiki/GettingStarted#up---dmlddl-updates

So, I argue that the directories not being sorted alphabetically on Linux is simply a bug, and we could skip the configuration altogether. Do you agree?

JscNZ:
Thanks @erikbra yeah to confirm those are directories (each containing upscripts). If you look at [DotNetfileSystemAccess](https://github.com/chucknorris/roundhouse/blob/master/product/roundhouse.core/infrastructure/filesystem/DotNetFileSystemAccess.cs#L395) class the methods like `get_all_file_name_strings_in(string directory, string pattern)` to get the filenames in a directory already order the script names - it's only retrieving the list of directories that seem to be affected.

I'm totally fine without a configuration flag and handling it as a bug. My only concern was that with this change if you in theory had used roundhouse to build a DB on prior version, then run that same set of scripts to build another DB you'd potentially get a different run order if you're running on linux.

erikbra:
Thanks for the clarification. I think it's not worth the effort/complexity to have a configuration flag on this, I see you point, but the "order" that the scripts are in in master now on Linux seems just random, and I don't think there is a big chance that anyone relied on this order. 

Could you please remove the configuration switch, and just always do

`Directory.GetDirectories(directory).OrderBy(d => d).ToArray():`

JscNZ:
Awesome it's done. You're totally right - thinking about it a little more having a configuration flag that's only used to correct a bug seems a bit backwards!

I just force pushed the branch 🙈 hopefully not an issue for you guys.

erikbra:
Force push ftw! (as long as it's not on master ;)) Thanks a lot for your contribution!

PawelBaranowski:
@erikbra Can we please release 1.2.1? #393 is essential for proper subdirectories handling on non-NTFS filesystems. #386 makes it impossible to switch to Linux for at least two people (@domingoladron and me) and I'm sure there are more.

erikbra:
Yes, sure, I can try to get it out. Sorry for the lack of response in general lately, there has been a lot of other stuff on my agenda too :/

PawelBaranowski:
You never know when life hits you. Tell me if I can be of any help.
BTW is there any discussion about current state of RH and future plans? Is there anything on the roadmap or is it in maintenance mode?

erikbra:
Thanks for your concern!

Good question on RH, really. I really want to move it forward, but I struggle a bit with no one else (almost) contributing. There are a few PRs here and there, which I appreciate a lot, but especially on some of the DB providers we are really lacking competence (especially Oracle). I want to move forward to .NET Core 3.1 (and soon-ish .NET 5), as the options for deployment are much better there (modern-style packaging as single file, e.g), but there are some technical dependencies we need to get rid of before going there (Log4Net, NHibernate). I should probably try to make some issues for each of these, to make it easier to separate the work

PR #394, opened 2019-10-16
Adding System.ValueTuple dependency

BlythMeister:
It would appear that this is now used as part of https://github.com/chucknorris/roundhouse/pull/388/files#diff-3139d803d9b04a25c7e3beaa80e76f4cR16
When consuming this in a framework project, this dependency is required


erikbra:
Thanks. Well spotted.

Issue #343, opened 2018-10-18
having issues getting connection string to work against Azure SQL


garthbf:
we are trying to run against Azure SQL that has Azure AD to auth. the Auth option of ActiveDirectoryIntegrated is givign an error.
We tried both ODBC and ADO connection string formats and the same error
Invalid value for key "authentication'
Running Command: .\RH.exe /c="Server=tcp:<servername>,1433;Database=,<Databasename>Uid=<Username>;Encrypt=yes;TrustServerCertificate=no;Connection Timeout=30;Authentication=ActiveDirectoryIntegrated" /f="mypath 
" /o="mypath\logs"  /dc /silent

Error:
Invalid value for key 'authentication'.
System.ArgumentException: Invalid value for key 'authentication'.
   at System.Data.Common.DbConnectionStringBuilderUtil.ConvertToAuthenticationType(String keyword, Object value)
   at System.Data.SqlClient.SqlConnectionStringBuilder.set_Item(String keyword, Object value)
   at System.Data.Common.DbConnectionStringBuilder.set_ConnectionString(String value)
   at System.Data.SqlClient.SqlConnectionStringBuilder..ctor(String connectionString)
   at roundhouse.databases.sqlserver.SqlServerDatabase.initialize_connections(ConfigurationPropertyHolder configuration_property_holder)
   at roundhouse.infrastructure.app.ApplicationConfiguraton.build_items_for_container(ConfigurationPropertyHolder configuration_property_holder)
   at roundhouse.infrastructure.app.ApplicationConfiguraton.build_the_container(ConfigurationPropertyHolder configuration_property_holder)
   at roundhouse.console.Program.set_up_configuration_and_build_the_container(String[] args, Mode mode)
   at roundhouse.console.Program.Main(String[] args)

erikbra:
First, I'll say that this is not supported, at the moment. We should look into whether we should let the Sql Connection itself decide what is a valid connection string or not, to be future-proof. Secondly, have you got this working with e.g. sqlcmd? Have you set up an ADFS server, etc, as describede here?

https://docs.microsoft.com/en-us/sql/connect/jdbc/connecting-using-azure-active-directory-authentication?view=sql-server-2017

garthbf:
we did get this to work. the issue was we had to change the connection string format to be csharp format.
you guys should document that somewhere. we had to get into the code to figure this out.

erikbra:
[...]

PR #395, opened 2019-10-18
Add Access Token support to SqlServerDatabase

### LINKED ISSUE #344: SqlConnectionStringBuilder: use of Access Token
Not currently supporting all of the connection options for sqlconnectionstringbuilder.  I would like to be able to consume an access token with the connection string.  There is no way currently to instantiate the access token?  

Example:
server_connection.AccessToken = accessToken;
server_connection.Open();

peckham:
Added support for access tokens for use in Azure SQL.

Unfortunately, I'm unable to test this against Azure. I did test against a local SQL Server instance and the property is set on the underlying connection.

Possibly fixes #344 and #343

### DIFF
--- a/product/roundhouse.console/Program.cs
+++ b/product/roundhouse.console/Program.cs
@@ -133,6 +133,9 @@ namespace roundhouse.console
                      string.Format(
                          "ConnectionStringAdministration - This is used for connecting to master when you may have a different uid and password than normal."),
                      option => configuration.ConnectionStringAdmin = option)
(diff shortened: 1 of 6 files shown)


erikbra:
Thank you very much for your contribution. Terribly sorry for the ridiculously long wait to have it merged.

kimjamia:
This was included in release 1.2.1, right? (It doesn't say it is but it looks like it is.)

PR #406, opened 2020-05-27
Add ability to push base64 encoded connection strings with password

qaswd:


erikbra:
Thank you for your contribution, @qaswd ! Sorry for the long delay before merging.

Issue #438, opened 2021-09-12
Going forward - request for input


erikbra:
# Background
Almost three years ago, I volunteered to take over the main maintainer responsibility for RoundhousE, as it was our weapon of choice for database migrations at my employer at the time. I'm really, really happy with the tool, and we still have over 5 years of database script history, that, using RoundhousE, can spin up a new database from 0 in a matter of seconds.

The simplicity of the tool is its elegance. No complicated ORMs, in-app migrations, etc. It speaks SQL, and it handles versioning of databases. And it does it well.

# Challenges

RoundhousE has been in development for over 10 years, the work was started by @ferventcoder , and he's done an amazing job. Time passing, though, some of the tools used in the beginning, and some of the external references from that time, makes it difficult to upgrade RoundhousE to current .net frameworks. In addition to this, the style originally used for writing tests, etc, BDD, is unfamiliar to me personally, and I find it difficult to expand and refactor the tool following some of the coding style, especially related to tests.

## Some problematic dependencies:
* Build pipeline and tools
    - Originally using an internal chucknorris tool, but not maintained
    - "Work around" with `build.ps1`, but could definitely be improved
    - AppVeyor used to build, but not actively maintained
    - No automatic push of built artifacts to NuGet and others
 * Third party libraries
     - log4net
     - NHibernate
     - Fody
     - Old versions of Microsoft Nuget packages
     - ILMerge
     - Oracle NuGet packages

## Lack of maintainers

Many people use RoundhousE, but few contribute. The challenges are especially related to maintaining some of the lesser used database providers, especially Oracle, which has some issues, but is on an ancient version of the oracle driver. But without active maintainers with knowledge of the Oracle integrations, it's difficult to upgrade

# Options going forward

My experience with maintaining RoundhousE (more or less successfully) for the last nearly three years, is that it's a bit difficult to maintain in its current state. I have plans to upgrade to .net 5 and .net 6, which brings new possibilities for e.g. distributing binaries that are single file and framework independent, so we could make distributions via e.g. apt-get and other package managers for Linux, and homebrew (or others) for macOS. However, in my opinion, that is difficult, given the current code base.

I have a couple of suggestions. Other suggestions are of course welcome - use the comments!

## Option 1 - Reboot RoundhousE

Re-write RoundhousE from scratch using modern .net. Keep external dependencies to a minimal, to make maintaining the project easier going forward. 

## Option 2 - Abandon RoundhousE - create a new project

[...]

PR #455, opened 2023-01-16
Update frameworks

Opvolger:
Hi,

I needed a update for the MySQL package. (bug fix). So I updated the project to net6.0 (3.1 = end of life).

Use your product, for a home-project.... (years, when it was still .net framework only).

Greetings Bas.


erikbra:
Hi, @Opvolger   - I've stopped maintaining RoundhousE, due to technical debt. I've created a re-implementation which aims to be 100% backwards compatible as well. Please give it a go, and see if it works in your use-case: https://erikbra.github.io/grate/

See https://github.com/chucknorris/roundhouse/issues/438#issuecomment-927371525 for more background on the issue.

PR #459, opened 2024-03-05
Chore: Added "abandoned" message to README.md

erikbra:
There are still some people that file issues or have questions on RoundhousE. I think the Internet all together is better off if we send them in the direction of grate, so we can concentrate our efforts there.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -1,3 +1,15 @@
+# NOTE: RoundhousE is "abandoned"
+
+RoundhousE is a wonderful tool, and I was one of the main maintainers for a couple of years, as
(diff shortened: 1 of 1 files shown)


erikbra:
@ferventcoder, @BiggerNoise: Are you OK with adding this notice on the Readme front page of RoundhousE? I think that all-in-all, it would be better to send people over to grate, and concentrate the joint efforts there. Do you agree?

BiggerNoise:
Yes, I think this makes sense. I'm unlikely to return to .NET anytime soon, so maintenance is likely to continue to mainly be your show.

